PR #66, opened 2014-10-20
[WIP/RFC] Rewrite things

adrienbrault:


stof:
@adrienbrault it would be great to open an issue to discuss the architecture change, to avoid loosing everything.
there are some interesting things in this refactoring, because the current architecture is based on a wrong composition: the ``ProcessorInterface`` gets decorated, but the ``sleep``, ``setDefaultOptions``, ``initialize`` and ``terminate`` method don't work with composition (making them unusable with the special StackedProcessor, which itself is broken if a processor decides to implement real composition for these methods)

PR #77, opened 2014-12-17
Add setters to PhpAmqpLibMessagePublisher

neeckeloo:
I add setters in order to use the same publisher instance to potentially publish messages to various channels and exchanges.


neeckeloo:
@stof I'm agree with you. It is not very relevant and unsafe in Swarrot itself to have a mutable publisher. So, I'm writing something in my side on top of Swarrot as you have advised me. Thanks.

PR #81, opened 2015-01-21
Keep message properties on retry processor

Domisys:
Actually, with retry processor, only the body of the message is keeped.
When the original message has properties such as "delivery_mode" or "app_id" the copied one loose those properties.

This PR fix that, keep the properties but override the "headers" one.

### DIFF
--- a/src/Swarrot/Processor/Doctrine/README.md
+++ b/src/Swarrot/Processor/Doctrine/README.md
@@ -1,6 +1,6 @@
 # ConnectionProcessor
 
-AckProcessor is a [swarrot](https://github.com/swarrot/swarrot) processor.
+ConnectionProcessor is a [swarrot](https://github.com/swarrot/swarrot) processor.
 
 The `doctrine_close_master` option is useful because otherwise once connected
 to the master the connection would never switch back to the slave.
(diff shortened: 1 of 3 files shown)


odolbeau:
Hi !
Thanks for your contribution... event if this subject is a bit... complicated. :/

In fact, if I remember well, at the start, all headers were send back to RabbitMQ. But @blaugueux found a bug and I opened [an issue on the php-amqp lib](https://github.com/pdezwart/php-amqp/issues/124).
After that, we remove all existing headers when sending back a message to Rabbit.

It seems you need some of them.
So what best solution? Send all headers except the `x-death` ? Or remove headers except `delivery_mode` and `app_id` ? I think the first one is the good solution (with a comment to explain why the `x-death` header is removed).
Is it OK for you @blaugueux?

Domisys:
We also catch the bug you issued.
In fact i just need the properties which couldn't contains arrays or nested arrays such. That's why i just keep all properties unless "headers" property.

Valid properties are:
content_type
content_encoding
priority
correlation_id
reply_to
expiration
message_id
timestamp
type
user_id
app_id
cluster_id

odolbeau:
-.-"

NEVER review a PR before taking a coffee.

PR #82, opened 2015-02-13
[RPC] Sending a message to a RPC consumer when it shouldn't

Taluu:
The properties `reply_to` and `correlation_id` are in fact always set, so the processor is always trying to send a new message back to an empty `reply_to` queue with an empty `correlation_id`

I added an empty check, but I'm not sure that `correlation_id` cannot have ld 0` value... or if I should test if it is strictly empty (`''`)


odolbeau:
I trust you. ^^

Taluu:
:D

PR #84, opened 2015-02-24
fix typo parameter name

L0rD59:
Reffer : https://github.com/swarrot/swarrot/blob/master/src/Swarrot/Processor/Retry/RetryProcessor.php#L64


odolbeau:
Thanks for the correction. :)

PR #86, opened 2015-04-14
Use Optonsresolver to be compatible with future versions

### LINKED ISSUE #85: The OptionsResolverInterface is deprecated as of Symfony 2.6
The typehint should be changed to rely on the OptionsResolver itself, for compatibility with future Symfony versions

odolbeau:
Close #85 

@stof we agree I just need to release a new patch version for this change?

### DIFF
--- a/src/Swarrot/Consumer.php
+++ b/src/Swarrot/Consumer.php
@@ -3,7 +3,6 @@
 namespace Swarrot;
 
 use Swarrot\Broker\MessageProvider\MessageProviderInterface;
-use Symfony\Component\OptionsResolver\OptionsResolverInterface;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 use Swarrot\Processor\ProcessorInterface;
 use Swarrot\Processor\ConfigurableInterface;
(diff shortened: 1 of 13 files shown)


stof:
@odolbeau changing the typehint in the interface is a BC break, so a patch version is wrong IMO.

Processors are the main extension point for the library, so these interfaces should be considered as part of the public API => major version bump.

odolbeau:
Seems logic. :) Thanks.

stof:
@odolbeau there was some interesting things in https://github.com/swarrot/swarrot/pull/66, even though it was not completed (the main point for me being the fact that it does not rely on a mix of composition and not-composition anymore but instead uses real composition for the ProcessorInterface, even though the implementation might not be the final one).
If you plan to bump the major version, it might be interesting to take a look at solving this issue. @adrienbrault are you still interested in this topic ?

odolbeau:
I agree, there is a lot of interesting things in #66.

Let's wait for @adrienbrault feedback before merging this one. :)

adrienbrault:
I'm not planning to work on this soon. But you can still bump the version and wait for interesting changes before actually releasing the new version right ?

stof:
@adrienbrault bumping the version without releasing would be a pain for other new features

ThisIsAreku:
I have a lot of warnings using Symfony 2.7.0. Do you plan to merge this PR soon ?

odolbeau:
v2.0.0 released. :)

PR #91, opened 2015-04-14
Correct SignalHandlerProcessor to avoid error when pcntl is not loaded

### LINKED ISSUE #87: the SignalHandlerProcessor does not deal properly with cases where pcntl is not available
On windows, pcntl is not available at all. The SignalHandlerProcessor tries to deal with it, but it does it only in ``process``. Other functions are broken:

- ``sleep`` should become a noop when the ext is not loaded instead of triggering a fatal error
- ``setDefaultOptions`` should avoid using the constants to define the default value (it should define an empty array in this case, but the option itself should still be registered so that it is accepted)

odolbeau:
Fix #87 

Is it ok for you @stof?

### DIFF
--- a/src/Swarrot/Processor/SignalHandler/SignalHandlerProcessor.php
+++ b/src/Swarrot/Processor/SignalHandler/SignalHandlerProcessor.php
@@ -42,7 +42,7 @@ class SignalHandlerProcessor implements ConfigurableInterface, SleepyInterface
     public function setDefaultOptions(OptionsResolverInterface $resolver)
     {
         $resolver->setDefaults(array(
-            'signal_handler_signals' => array(SIGTERM, SIGINT, SIGQUIT)
+            'signal_handler_signals' => extension_loaded('pcntl') ? [SIGTERM, SIGINT, SIGQUIT] : []
         ));
     }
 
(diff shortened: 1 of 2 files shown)


src/Swarrot/Processor/SignalHandler/SignalHandlerProcessor.php
```diff
@@ -41,6 +41,10 @@ public function __construct(ProcessorInterface $processor, LoggerInterface $logg
      */
     public function setDefaultOptions(OptionsResolverInterface $resolver)
     {
+        if (!extension_loaded('pcntl')) {
+            return;
```
stof: this is not. The ``signal_handler_signals`` option should still be know by the OptionsResolver to prevent exceptions when using it because of an unknown option

src/Swarrot/Processor/SignalHandler/SignalHandlerProcessor.php
```diff
@@ -51,6 +55,10 @@ public function setDefaultOptions(OptionsResolverInterface $resolver)
      */
     public function sleep(array $options)
     {
+        if (!extension_loaded('pcntl')) {
+            return false;
```
stof: this should return ``true``. Otherwise, you always stop it when the queue is empty (which is the job of the InsomniacProcessor)

stof:
Hurrah. HHVM broke their phar support...

odolbeau:
@stof Is there something to do to avoid the error in hhvm tests? It's also broken on master.
Should I update the `.travis.yml` to accept failures with hhvm?

stof:
@odolbeau I would do that for now yes, until they fix this

PR #92, opened 2015-04-15
Implement the MemoryLimitProcessor

### LINKED ISSUE #89: Add a memory limit processor
This processor would stop the consumer when the memory usage gets too close from the PHP memory limit, to provide a graceful shutdown (with logging) in such case instead of waiting until the memory limit is reached and the process crashes. This would help in case your processor is leaking memory.

what do you think @odolbeau ?

stof:
Closes #89

Currently, the default value is to be unlimited, because I don't know what a good generic default could be.

Note that the intended usage is to register it outside the ObjectManagerProcessor (because clearing the entity manager is likely to release memory by deleting references to the entities), or any other processor doing a cleanupof the memory after the processing. The goal is to limit the memory to be leaked after processing (if the processing of a single message reaches the memory limit because of its own needs, there is nothing we can do to intercept it anyway, as we can only hook before or after the logic, not inside it)

### DIFF
--- a/src/Swarrot/Processor/MemoryLimit/MemoryLimitProcessor.php
+++ b/src/Swarrot/Processor/MemoryLimit/MemoryLimitProcessor.php
@@ -1,0 +1,66 @@
+<?php
+
+namespace Swarrot\Processor\MemoryLimit;
+
+use Psr\Log\LoggerInterface;
+use Swarrot\Broker\Message;
+use Swarrot\Processor\ConfigurableInterface;
+use Swarrot\Processor\ProcessorInterface;
+use Symfony\Component\OptionsResolver\OptionsResolverInterface;
+
+class MemoryLimitProcessor implements ConfigurableInterface
+{
+    /**
+     * @var ProcessorInterface
+     */
+    private $processor;
+
+    /**
+     * @var LoggerInterface
+     */
+    private $logger;
+
+    /**
+     * @param ProcessorInterface $processor Processor
+     * @param LoggerInterface    $logger    Logger
+     */
+    public function __construct(ProcessorInterface $processor, LoggerInterface $logger = null)
+    {
+        $this->processor = $processor;
+        $this->logger    = $logger;
+    }
+    /**
+     * {@inheritDoc}
+     */
+    public function process(Message $message, array $options)
+    {
+        $return = $this->processor->process($message, $options);
+
+        if (null !== $options['memory_limit'] && memory_get_usage() >= $options['memory_limit'] * 1024 * 1024) {
+            $this->logger and $this->logger->info(
+                sprintf('[MemoryLimit] Memory limit has been reached (%d MB)', $options['memory_limit']),
+                [
+                    'swarrot_processor' => 'memory_limit'
+                ]
+            );
+
+            return false;
+        }
+
+        return $return;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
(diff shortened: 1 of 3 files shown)


odolbeau:
Isn't `ini_get('memory_limit')` enough?

src/Swarrot/Processor/MemoryLimit/MemoryLimitProcessor.php
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * @param ProcessorInterface $processor Processor
+     * @param LoggerInterface    $logger    Logger
+     */
+    public function __construct(ProcessorInterface $processor, LoggerInterface $logger = null)
+    {
+        $this->processor = $processor;
+        $this->logger    = $logger;
+    }
+    /**
```
odolbeau: Missing blank line

stof:
@odolbeau if the PHP memory_limit is reached, it is too late (fatal error already happened). And even if we use 95% of this limit, it might also be too late, given that this concerns only the memory after the processing, so the one not cleaned between messages

tests/Swarrot/Processor/MemoryLimit/MemoryLimitProcessorTest.php
```diff
@@ -0,0 +1,50 @@
...
+use Swarrot\Broker\Message;
+
+class MemoryLimitProcessorTest extends ProphecyTestCase
+{
+    public function test_it_is_initializable_without_a_logger()
+    {
+        $processor = $this->prophesize('Swarrot\Processor\ProcessorInterface');
+
+        $processor = new MemoryLimitProcessor($processor->reveal());
+        $this->assertInstanceOf('Swarrot\Processor\MaxMessages\MemoryLimitProcessor', $processor);
```
odolbeau: Wrong namespace

tests/Swarrot/Processor/MemoryLimit/MemoryLimitProcessorTest.php
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    public function test_it_is_initializable_with_a_logger()
+    {
+        $processor = $this->prophesize('Swarrot\Processor\ProcessorInterface');
+        $logger    = $this->prophesize('Psr\Log\LoggerInterface');
+
+        $processor = new MemoryLimitProcessor($processor->reveal(), $logger->reveal());
+        $this->assertInstanceOf('Swarrot\Processor\MaxMessages\MemoryLimitProcessor', $processor);
```
odolbeau: Wrong namespace

stof: fixed

rcrowe:
Any chance we can get this merged in?

stof:
Well, I haven't continued the work on it, because it is not very useful by itself: it does not prevent crashes if you have a processor consuming too much memory (it is too late).
The only thing it could be used for would be to kill the worker in case it leaks memory (i.e. it still uses lots of memory after processing messages). And this is not really a use case I need (my workers are not leaking like that)

rcrowe:
We do for a few jobs & it's helpful to kill before getting to high. This looks complete & wonder what stops it getting merged?

odolbeau:
I know the maintainer of this project and it's probably because he can be a bit lazy. ;)

Btw, thanks @stof for this processor. :)

PR #94, opened 2015-06-03
Fixed deprecated on setAllowedTypes

ThisIsAreku:
This PR complement #86 and remove remaining warnings


odolbeau:
Thanks. I created a new version too quickly. :/
Once travis is OK for master, I create a v2.0.1.

stof:
This does not support Symfony 2.5 and lower anymore. You should keep support of the 2.3 LTS IMO by falling back to the old API when needed

PR #95, opened 2015-06-03
Add testing against the lowest dependencies on Travis

stof:
The goal is to ensure that the lowest bound is set properly, and avoid issues like in https://github.com/swarrot/swarrot/pull/94 breaking the support of older versions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ matrix:
     include:
         # Test against librabbitmq cutting-edge
         - { php: 5.5, env: LIBRABBITMQ_VERSION=master }
+        - { php: 5.4, env: COMPOSER_FLAGS='--prefer-lowest --prefer-stable' }
     allow_failures:
         - php: hhvm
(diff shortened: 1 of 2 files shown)


odolbeau:
:+1: 
I merge this PR and I update the lib to avoid this error.
Thanks BTW

stof:
@odolbeau I was already fixing some lowest bound being wrong (the fact that Travis is slowed down currently makes this process longer, and I don't have the testing environment set locally for it).

odolbeau:
Cool. :)
In this case, I let you finish. #lazy ;)

stof:
@odolbeau fixing the composer.json for lowest deps is actually done. The only remaining work is fixing the change done in #94 to avoid dropping the support of the Symfony 2.3 LTS (which is a bit too early IMO as the new 2.7 LTS still has a bunch of issues in 2.7.0, and 2.3 is still supported for 1 year)

PR #98, opened 2015-06-04
Fix compatibility with OptionsResolver <2.6

stof:
https://github.com/swarrot/swarrot/pull/94 broke the compatibility with the 2.3 LTS version by using the API introduced in Symfony 2.6.
This adds support back for 2.3 by detecting whether the new API is available or no (based on the existence of another method of the new API as it is faster and easier than using reflection to detect the number of arguments in the method)

### DIFF
--- a/src/Swarrot/Processor/Ack/AckProcessor.php
+++ b/src/Swarrot/Processor/Ack/AckProcessor.php
@@ -85,6 +85,13 @@ class AckProcessor implements ConfigurableInterface
             'requeue_on_error' => false
         ));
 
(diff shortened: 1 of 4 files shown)


stof:
See #99 for the remaining issue for the lowest deps

odolbeau:
:+1: 
Thanks a lot. :)

ThisIsAreku:
@odolbeau can you please tag a new version for https://github.com/swarrot/swarrotBundle with the latest version of Swarrot ? Thanks :)

stof:
SwarrotBundle does not need any change for that though.

PR #100, opened 2015-06-04
Run php-cs-fixer

odolbeau:


src/Swarrot/Broker/Message.php
```diff
@@ -22,7 +22,7 @@ class Message
     protected $id;
 
     /**
-     * __construct
+     * __construct.
```
stof: this short description is not very useful though :)

src/Swarrot/Broker/MessageProvider/MessageProviderInterface.php
```diff
@@ -7,33 +7,29 @@
 interface MessageProviderInterface
 {
     /**
-     * get
+     * get.
```
stof: IMO, you could remove such useless short description (or provide a useful one)

odolbeau:
Yes. I really need to spend some time to update all useless PHPDoc and add a real doc instead of some readme files. :)

PR #102, opened 2015-06-17
Fix #101: Don't set default values for non-existing properties in PhpAmqpLibLMessageProvider

### LINKED ISSUE #101: Retry Processor failing with php-amqplib
Hi,

When I use Swarrot (v2.0.2) (or rather SwarrotBundle (dev-master), but I believe the problem is in Swarrot) with the retry processor, I get the following error when that processor tries to republish a message to the retry exchange:
```
  [PhpAmqpLib\Exception\AMQPProtocolChannelException]                                      
  PRECONDITION_FAILED - user_id property set to '0' but authenticated user was '******'  
```
I think the problem may have been introduced by https://github.com/swarrot/swarrot/pull/62 as it is setting default values conflicting with the way amqplib (v2.5.1) works.
Fixing the problem might be as easy as setting keys only for properties that actually have a value. 

If it makes sense, I can try to submit a PR.

jean-gui:


odolbeau:
Thanks @jean-gui!

@astorije what do you think of this PR?
I think it conflict with #62. Do you have another idea to fix #101?

Maybe it's possible to remove default values from headers in the retry processor? Even if it's probably not very elegant. :/

jean-gui:
@odolbeau Using something like the following code in the retry processor seems to do the job as well:
```php
foreach ($properties as $key => $value) {
    if ($value == null) {
        unset($properties[$key]);
    }
}
```
But it feels like it shouldn't be fixed in that processor as other processors could run into the same issue.

jean-gui:
@odolbeau Any news on that ? If you are awaiting an answer from @astorije, he is a former colleague of mine and used Swarrot only to prototype an app. I doubt he still has an interest in it.

Taluu:
I guess that `array_filter` should be your friend here to remove null properties. Buit this seems legit (cc @odolbeau)

odolbeau:
Thanks @Taluu! :)
@jean-gui can you use an array_filter please? Then, I'll merge your PR. :)

jean-gui:
@odolbeau @Taluu You mean you prefer https://github.com/swarrot/swarrot/pull/102#issuecomment-112810263 (with array_filter) over the current pull request?

Taluu:
Mmh now that I'm looking at that, I guess this manipulation should be a no-go (the thing for the RetryProcessor), because as you said it, it may collide with other processors... You may forget the `array_filter` (the possibility to filter on the key values is only from php 5.6 :()

odolbeau:
So, I guess I can merge. :)

Thanks for you code review @Taluu.

Sorry for the delay @jean-gui but as I don't use the `PhpAmqpLibLMessageProvider` myself, I prefer to ask confirmation to someone who use it.

Taluu:
It's not because I made the publisher that I'm using it :D

jean-gui:
@odolbeau no problem and thanks for merging :)

PR #104, opened 2015-06-29
Allow to infinitely requeue with Retry processor

jean-gui:
When using the retry processor, I thought it could be useful to requeue messages even if `retry_attempts` was reached rather than losing them.

A new option was added, `keep_retrying`, that when set to true (defaults to false) will requeue failed messages with `%attempt%` in the routing key set to `retry_attempts` option, and will keep incrementing the header `swarrot_retry_attempts`.

### DIFF
--- a/src/Swarrot/Processor/Retry/README.md
+++ b/src/Swarrot/Processor/Retry/README.md
@@ -25,8 +25,8 @@ the routing_key `key_1`.
 
 If `retry_final_key` is set and `retry_attempts` is reached, the message will be 
 published to the exchange `retry` with the routing key `retry_final_key`.
-If `retry_final_key` is not set and `retry_attempts` is reached, the message will be 
-lost and an exception thrown.
+If `retry_final_key` is not set and `retry_attempts` is reached an exception will
+be thrown.
 
 ## Real example
 


src/Swarrot/Processor/Retry/RetryProcessor.php
```diff
@@ -61,7 +61,11 @@ public function process(Message $message, array $options)
                 $properties
             );
 
-            $key = str_replace('%attempt%', $attempts, $options['retry_key_pattern']);
+            $key = str_replace(
+                '%attempt%',
+                ($attempts > $options['retry_attempts']) ? $options['retry_attempts'] : $attempts,
```
odolbeau: If I understand well, once the retry_attempts is reached, it will republish in the routing_key as before?
Not sure it's very understandable if the processor continue to publish in `something_retry_3` (for example) after 10 attempts. :/
In your use case is it possible to have a very high attempts limit and a routing_key which doesn't contains %retry%?

jean-gui: That would most certainly be possible and seems more logical indeed. What about an option that when with a specific routing key would use that key instead of `retry_key_pattern`? I can't think of a good name though. What about `retry_final_key`?
I think that `keep_retrying` would then become unnecessary.

odolbeau: So for a given number of retries you want to publish with specific routing_keys (`smtg_retry_1`, `smtg_retry_2`, `smtg_retry_3`) and once the attempt limit is reached you want to always publish with a dedicated routing_key (`smtg_retry`)?

jean-gui: Yes.

jean-gui:
I've updated the MR using the previously discussed final routing key rather than reusing smtg_retry_3. Does this seem reasonable?

It seems that Travis is having issues with composer, I don't know if you can manually trigger a new build.

I will also have to make a MR to swarrot-bundle to add the new option there. Should I wait until this one is merged?

src/Swarrot/Processor/Retry/README.md
```diff
@@ -22,6 +23,11 @@ in the exchange named `retry` and if the `retry_key_pattern` is
 publish a new message (similar to the first one) in the exchange `retry` with
 the routing_key `key_1`.
 
+If `retry_final_key` is set and `retry_attempts` is reached, the message will be 
+published to the exchange `retry` with the routing key `retry_final_key`.
+If `retry_final_key` is not set and `retry_attempts` is reached, the message will be
```
odolbeau: Can you remove `the message will be lost` please?
It could be true but not always. It depends (a lot) on your own configuration. :)

odolbeau:
Build restarted. :)

Anyway, I think you can reach your goal by using 2 RetryProcessor:
* the first one with a given `retry_attempt` and a pattern containing `%attempt%`
* the second one with a static pattern and infinite attempts.

This way, the RetryProcessor stay really simple as the only thing to add is the ability to make infinite attempts.

What do you think about this solution?

jean-gui:
Creating a new Processor could indeed be a better way to solve that issue. I guess it could be called `RequeueProcessor` and take one parameter, `requeue_routing_key`.
It could have its own counting header (`swarrot_requeue_count`).

This processor should also drop the header set by RetryProcessor. Are there any other headers it should drop?

odolbeau:
Sorry for the delay, I'm pretty busy these days... :/

Instead of creating a new Processor, I think you can reuse the same one.
I was thinking about something like this:

```
    ->push('Swarrot\Processor\Retry\RetryProcessor', ['retry_attempts' => -1, 'retry_key_pattern' => 'infinite'])
    ->push('Swarrot\Processor\Retry\RetryProcessor')
;
```

I this case the only missing feature is to support infinite retry when `retry_attempts == -1`.

What do you think about this solution?

Taluu:
:+1:

jean-gui:
Ah OK, I had misunderstood your comment about two processors.
My concern with that proposal is about SwarrotBundle. Will it be able to handle two RetryProcessors with different parameters?

odolbeau:
You're right, it's not possible. :/
This bundle really need some love. :/

Maybe you can create your own Processor which extends the current `RetryProcessor` and changed the default options?

PR #105, opened 2015-07-16
Add a workaround for the bug in pecl amqp not exposing the delivery mode

### LINKED ISSUE #103: the retry processor fails saying that the processing mode is unknown
When queueing a message for retry, the processor fails:

```
Server channel error: 406, message: PRECONDITION_FAILED - invalid message: {delivery_mode_unknown,0}
```

I'm using the pecl extension with ext-amqp 1.6.0beta2 (compiled against librabbitmq 0.5.2) and RabbitMQ 3.4.2

stof:
Closes #103 by implementing a workaround until the PECL extension is fixed (if we don't know the original delivery mode, we will just let RabbitMQ use the default one)


odolbeau:
Sounds good for me! Thanks!

PR #107, opened 2015-08-15
Add options usage example

kyar:
Options usage is not obvioulsy without example


odolbeau:
I really need to work on the documentation. :/

PR #111, opened 2015-10-20
On retry keep headers except nested arrays

Domisys:
Actually, with retry processor, original message headers are removed to prevent issue described here : https://github.com/pdezwart/php-amqp/issues/124.

This PR allows to keep message headers except nested arrays.

### DIFF
--- a/src/Swarrot/Processor/Retry/RetryProcessor.php
+++ b/src/Swarrot/Processor/Retry/RetryProcessor.php
@@ -52,9 +52,14 @@ class RetryProcessor implements ConfigurableInterface
                 throw $e;
             }
 
-            $properties['headers'] = array(
-                'swarrot_retry_attempts' => $attempts,
-            );
+            // workaround for https://github.com/pdezwart/php-amqp/issues/124
+            if (isset($properties['headers'])) {
+                $properties['headers'] = array_filter($properties['headers'], function($headerValue) {
+                    return ! is_array($headerValue);
+                });
+            }
+            
+            $properties['headers']['swarrot_retry_attempts'] = $attempts;
 
             // workaround for https://github.com/pdezwart/php-amqp/issues/170. See https://github.com/swarrot/swarrot/issues/103
             if (isset($properties['delivery_mode']) && 0 === $properties['delivery_mode']) {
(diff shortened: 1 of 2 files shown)


stof:
Instead of putting the workaround in the RetryProcessor, it would be better to put it at the publisher level (and then, you can avoid applying the workaround in ext-amqp >= 1.6.0beta2, as the issue has been fixed)

Domisys:
Hi,

Actually, this is the RetryProcessor that drops message original headers. So I think this fix (until php-amqp is in beta) has to be done in this processor.

So instead of dropping all original headers, we do it less "agressively" by only removing those with nested arrays.

**More context:**
Actually, when consuming a message, we need (during message processing) some data contained in message header. But when this message is republished by retryProcessor ,all headers are dropped except *swarrot_retry_attempts*. Apparently, this is done because of "nested arrays" issue on php-amqp (as explained on this [PR](https://github.com/swarrot/swarrot/pull/81#issuecomment-70986520)). But normally, we should not lose these headers.

stof:
@Domisys what I mean is that RetryProcessor should keep *all* headers, and that the publisher should contain the necessary workaround (dropping the nested arrays). this would apply the workaround only for the publisher needing it (and it can detect the ext-amqp version to decide whether it is needed or no)

Domisys:
Ok, sorry I didn't understand your previous comment.

As [previous fixes](https://github.com/swarrot/swarrot/commit/937c82648c31846a4acee47da24c9615f812f210) were done in this processor, I thought it was appropriate to produce the fix here.

@odolbeau, what is your opinion ?

stof:
Well, IMO, previous fixes were done in the wrong place

odolbeau:
I agree with @stof, as we know there is a problem publishing headers containing arrays, the `MessagePublisher` should handle it.

In fact, the other workaround in the `RetryProcessor` (when `delivery_mode = 0`) should also be moved in the `MessagePublisher`.

These workarounds are specific to only 1 implementation (pecl extension). If someone wants to use swarrot with SQS for example (even if there is no implementation for now), removing nested arrays doesn't make sense. That's why it should be handle by the right `MessagePublisher`.

PR #113, opened 2015-11-10
Exception handling is now in a function.

remyLemeunier:
Here is my problem: I want to retry messages only on some exceptions.
For instance, if it's a network problem i want to retry but for others i don't.
Some providers charges you on each queries, it's not worth it to retry on known exceptions.

Currently we can't extend the RetryProcessor, because the whole code is located in the function process.

I have moved the exception handling part in another function.
This way, i can now extend the RetryProcessor and do my own stuff, and then call this function if i want to retry.

What do you think?

### DIFF
--- a/src/Swarrot/Processor/Retry/RetryProcessor.php
+++ b/src/Swarrot/Processor/Retry/RetryProcessor.php
@@ -23,69 +23,73 @@ class RetryProcessor implements ConfigurableInterface
     }
 
     /**
-     * {@inheritDoc}
+     * {@inheritdoc}
      */
     public function process(Message $message, array $options)
     {
         try {
             return $this->processor->process($message, $options);
         } catch (\Exception $e) {
-            $properties = $message->getProperties();
-
-            $attempts = 0;
-            if (isset($properties['headers']['swarrot_retry_attempts'])) {
-                $attempts = $properties['headers']['swarrot_retry_attempts'];
-            }
-            $attempts++;
-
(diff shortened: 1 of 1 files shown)


odolbeau:
Can you rebase your branch please?

src/Swarrot/Processor/Retry/RetryProcessor.php
```diff
@@ -23,69 +23,73 @@ public function __construct(ProcessorInterface $processor, MessagePublisherInter
...
 
+    /**
+     * Retry on exception.
+     * Retries are stopped once retry_attempts has been reached.
+     *
+     * @param \Exception $e
+     * @param Message    $message
+     * @param array      $options
+     */
+    protected function onProcessorException(\Exception $e, Message $message, array $options)
```
stof: if you want to handle Throwable on PHP 7 as suggested in #129, the ``\Exception`` typehint should be removed. Otherwise, it is impossible to update the method signature later (changing the signature of a protected method is a BC break)

stof: and I would name the method ``handleException`` rather than ``onProcessorException``, to have a verb in the name

odolbeau: Fortunately, this method don't exist yet, no need for a new major version here. :)
I agree with the name of the method + the handling of Throwable. :)

odolbeau:
Solved by #129

PR #118, opened 2015-12-26
Get rid of phpspec/prophecy-phpunit

odolbeau:


composer.json
```diff
@@ -16,9 +16,7 @@
         "psr/log": "~1.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "~4.0",
-        "phpspec/prophecy-phpunit": "~1.0",
-        "phpspec/prophecy": "~1.4",
+        "phpunit/phpunit": "~4.7",
```
stof: you should also allow PHPUnit 5

PR #121, opened 2016-01-12
Change log level from warning to error

marcw:
This is a proposal to change the log level from `warning` to `error` in the `AckProcessor` processor. 

I changed it in order to be more in line with the semantics of the AckProcessor. In my mind, exceptions are errors and should be logged as this. 

Plus, the PSR Log Interface gives us some hint on the right log level we should use: 
- [error level doc](https://github.com/php-fig/log/blob/master/Psr/Log/LoggerInterface.php#L57-L66)
- [warning level doc](https://github.com/php-fig/log/blob/master/Psr/Log/LoggerInterface.php#L68-L79)


odolbeau:
Makes sense. :)
Thanks!

PR #123, opened 2016-02-12
videlalvaro/php-amqplib is abandonned, moving to php-amqplib/php-amqplib ?

olaurendeau:
Got this installing swarrot
capture d ecran 2016-02-12 a 12 17 29


odolbeau:
:+1:

PR #124, opened 2016-02-12
Is php7 supported ?

olaurendeau:


odolbeau:
Let's try. :)

odolbeau:
Seems good. :)

PR #125, opened 2016-03-30
PHPDoc updated (description + corrected @return param)

mfix:
As discusses some time ago - @return statement from ProcessorInterface was not complete. I also updated description a bit - hope it will be helpful for someone.


odolbeau:
Thanks! :+1:

PR #126, opened 2016-04-19
Propagating interfaces to decorated processors

Taluu:
On the presented processors in Swarrot, if the decorated processors (design used in almost every processors that are built-in with Swarrot) implements the interfaces that adds things such as options, sleep possibilities, ... etc.

I also added the `ProcessorInterface`, but this is not really in the scope of this PR though (IMO, the interfaces should not extend the `ProcessorInterface`, it does not really make sense). But this is only my opinion though.

### DIFF
--- a/src/Swarrot/Processor/Ack/AckProcessor.php
+++ b/src/Swarrot/Processor/Ack/AckProcessor.php
@@ -4,17 +4,20 @@ namespace Swarrot\Processor\Ack;
 
 use Swarrot\Processor\ProcessorInterface;
 use Swarrot\Processor\ConfigurableInterface;
+use Swarrot\Processor\InitializableInterface;
+use Swarrot\Processor\DecoratorTrait;
+use Swarrot\Processor\SleepyInterface;
+use Swarrot\Processor\TerminableInterface;
 use Swarrot\Broker\MessageProvider\MessageProviderInterface;
 use Swarrot\Broker\Message;
 use Psr\Log\LoggerInterface;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 
-class AckProcessor implements ConfigurableInterface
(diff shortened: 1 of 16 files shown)


stof:
-1 for this approach, as it makes them incompatible with the StackedProcessor (logic will be called both in this case), and it makes writing processors much harder (you will always have to implement all interfaces and take care of the propagation).
thus, even your implementation is broken (sleep is meant to return a value for instance, so your trait is broken for instance).

There was a suggestion for a big refactoring in https://github.com/swarrot/swarrot/pull/66 which was not completed, but which aimed at cleaning the architecture. your proposal only makes it even more messy.

Taluu:
You actually don't have to implement everything if your processor does not decorate another one. But you have to do it if you have one though

Taluu:
Note : We are currently blocked and can't use any of the mentionned processors (such as the `MemoryLimitProcessor`, as we need the options in our sub processors to make it work.

But, the options are not declared, and using a stack processor with a main processor that _also_ declares our own processor is IMO a huge flaw :

```php
<?php

$processor = new MemoryProcessor(new MyProc);
$stack = new StackProcessor($processor, [new MyProc]);
```

instanciating a processor just for the heck of declaring options (and sleep, etc), especially as some may depend on things that happens in the `process` method (such as the RPC ones). So the `stack` processor is basically unpraticable.

And yes I guess I have a better handling of the sleep method in the trait, I'll take care of that in a moment

stof:
@Taluu use the Stackbuilder.  This is the way things are expected to be used today.

odolbeau:
I agree with @stof, the StackBuilder should make the job correctly. :)

I really have to take time to give some love to the #66 refactoring proposal which makes things cleaner IMO.

PR #129, opened 2016-05-30
Catch PHP7 Throwable

MattKetmo:
Catching Throwable helps to catch a larger range of errors since PHP 7 (and avoid a worker to break while it's running).

Here I duplicated the `catch \Exception` block in order to keep the compatibility with PHP 5.

Note that this logic must be applied in other places:
- AckProcessor
- InstantRetryProcessor
- NewRelicProcessor
- RetryProcessor (need merging #113 first)

### DIFF
--- a/src/Swarrot/Processor/Ack/AckProcessor.php
+++ b/src/Swarrot/Processor/Ack/AckProcessor.php
@@ -55,23 +55,10 @@ class AckProcessor implements ConfigurableInterface
             );
 
             return $return;
+        } catch (\Throwable $e) {
+            $this->handleException($e, $message, $options);
         } catch (\Exception $e) {
-            $requeue = isset($options['requeue_on_error']) ? (boolean) $options['requeue_on_error'] : false;
-            $this->messageProvider->nack($message, $requeue);
-
-            $this->logger and $this->logger->error(
-                sprintf(
-                    '[Ack] An exception occurred. Message #%d has been %s.',
-                    $message->getId(),
-                    $requeue ? 'requeued' : 'nack\'ed'
-                ),
-                [
-                    'swarrot_processor' => 'ack',
-                    'exception' => $e,
-                ]
-            );
-
-            throw $e;
+            $this->handleException($e, $message, $options);
         }
     }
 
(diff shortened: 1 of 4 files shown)


stof:
Looks like a good idea. It would be great to update other processors.

regarding #113, you might want to take it over (while preserving commit authorship) as it looks stalled.

odolbeau:
Thank you @MattKetmo! <3
I can't wait to see this PR finished! ;)

MattKetmo:
I moved the code in the catch in a separate method `handleException` for each processor to simplify the duplication of the code inside.

Note that this method is private and thus does not solve #113 (intentionally). Extending such a method is rarely a good choice. Better would be to provide another (simpler) method to override or an option to check the exception type.

src/Swarrot/Processor/Ack/AckProcessor.php
```diff
@@ -93,4 +80,29 @@ public function setDefaultOptions(OptionsResolver $resolver)
             ));
         }
     }
+
+    /**
+     * @param Throwable|Exception  $exception
```
stof: wrong types. Missing ``\`` to make them FQCN rather than relative class names

src/Swarrot/Processor/NewRelic/NewRelicProcessor.php
```diff
@@ -43,6 +43,13 @@ public function process(Message $message, array $options)
 
         try {
             $result = $this->processor->process($message, $options);
+        } catch (\Throwable $e) {
+            if ($this->extensionLoaded) {
+                newrelic_notice_error(null, $e);
```
stof: do they handle PHP 7 Error properly ?

src/Swarrot/Processor/Ack/AckProcessor.php
MattKetmo: Fixed

src/Swarrot/Processor/NewRelic/NewRelicProcessor.php
MattKetmo: Good question. I've reverted this change as a precaution.

MattKetmo:
I've cleaned up a bit the composer.json (replace `~` with `^` and switch to PSR-4). Feel free to keep the commit or remove it ;)

Taluu:
It's not really psr-4 though, but still, :+1: :D

MattKetmo:
PSR-4 is not (only) about moving `src/Swarrot/*` to `src/*`. PSR-0 is simply deprecated and should not be used anymore.

Taluu:
Hence my :+1: :}

Taluu:
But to be honest, I'm not sure this is in the scope of this PR. Otherwise, you should also replace the `|` by `||` in version constraints as this is the official version now.

But I really think this should be done in a separate PR (moving or not the files, as that's besides the point)

MattKetmo:
Composer changes move to #130

odolbeau:
Looks good for me!
Thanks for your review @stof & @Taluu! :)

BTW thanks @MattKetmo! :)

PR #130, opened 2016-05-31
Cleanup composer.json

MattKetmo:
- Move from `~` to `^`
- Replace `|` by `||`
- Use PSR-4

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name":        "swarrot/swarrot",
+    "name": "swarrot/swarrot",
     "description": "A simple lib to consume RabbitMQ queues",
     "keywords": ["swarrot","worker","queue","amqp"],
     "type": "library",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -11,27 +11,27 @@
...
+        "php": "^5.4 || ^7.0",
+        "symfony/options-resolver": "^2.3 || ^3.0",
+        "psr/log": "^1.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "~4.7|~5.0",
-        "php-amqplib/php-amqplib": "~2.1",
-        "doctrine/common": "~2.3",
-        "doctrine/dbal": "~2.2"
+        "phpunit/phpunit": "^4.7 || ^5.0",
```
Taluu: I wonder if we should not remove the phpunit dependency, and use the global one on travis instead ? Or even suggest using the global one if we need to run the tests...

MattKetmo: Travis tests already uses its global one.
But personally I don't have phpunit installed globally anymore :)

odolbeau: Same for me. ;)

Taluu: I actually did the reverse, I installed it globally not too long ago. :D

MattKetmo: now you can run `composer test` instead of `phpunit` ;)

odolbeau:
Thanks a lot! <3

PR #133, opened 2016-06-16
[SignalHandlerProcessor] use initialize to register signals handlers

krichprollsch:
the register must be done at the very first start of the processor not after a first message processed


odolbeau:
Currently it's done before the processing of the first message. As the goal of this processor is to avoid any interruption when a message is processing, does it makes sense to start it before?

I mean, if you launch a worker on an empty queue, you don't need to catch signals as the worker can be stopped at any moment.
On the other hand it seems more logic to start it at the beginning and it force the worker to finish properly.

At the end, thanks for your PR. :)

krichprollsch:
hum good point, you are right. But this way avoid the register part after each message process.
And you get the stop message all the time as expected.

Thank you for the merge!

PR #134, opened 2016-09-23
add sqs provider

nicolasThal:
I will share the provider I have written to use swarrot with amazon sqs.

This code is already in production on our platform.

As we are currently changing our queueing system, we didn't use a message publisher yet

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,13 +20,13 @@
         "php-amqplib/php-amqplib": "^2.1",
         "doctrine/common": "^2.3",
         "doctrine/dbal": "^2.2",
-        "aws/aws-sdk-php": "2.*"
+        "aws/aws-sdk-php": "^2.8 || ^3.0"
     },
     "suggest": {
         "pecl-amqp": "*",
         "php-amqplib/php-amqplib": "^2.1",
         "symfony/console": "^2.4 || ^3.0",
-        "aws/aws-sdk-php": "2.*"
+        "aws/aws-sdk-php": "^2.8 || ^3.0"
     },
     "autoload": {
         "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -19,12 +19,14 @@
         "phpunit/phpunit": "^4.7 || ^5.0",
         "php-amqplib/php-amqplib": "^2.1",
         "doctrine/common": "^2.3",
-        "doctrine/dbal": "^2.2"
+        "doctrine/dbal": "^2.2",
+        "aws/aws-sdk-php": "2.*"
     },
     "suggest": {
         "pecl-amqp": "*",
         "php-amqplib/php-amqplib": "^2.1",
-        "symfony/console": "^2.4 || ^3.0"
+        "symfony/console": "^2.4 || ^3.0",
+        "aws/aws-sdk-php": "2.*"
     },
     "autoload": {
         "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -20,13 +20,13 @@
         "php-amqplib/php-amqplib": "^2.1",
         "doctrine/common": "^2.3",
         "doctrine/dbal": "^2.2",
-        "aws/aws-sdk-php": "^2.8 || ^3.0"
+        "aws/aws-sdk-php": "^2.8"
     },
     "suggest": {
         "pecl-amqp": "*",
         "php-amqplib/php-amqplib": "^2.1",
         "symfony/console": "^2.4 || ^3.0",
-        "aws/aws-sdk-php": "^2.8 || ^3.0"
+        "aws/aws-sdk-php": "^2.8"
     },
     "autoload": {
         "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         "psr/log": "^1.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "^4.7 || ^5.0",
+        "phpunit/phpunit": "^4.8.27 || ^5.0",
         "php-amqplib/php-amqplib": "^2.1",
         "doctrine/common": "^2.3",
         "doctrine/dbal": "^2.2",
--- a/src/Swarrot/Broker/MessageProvider/SqsMessageProvider.php
+++ b/src/Swarrot/Broker/MessageProvider/SqsMessageProvider.php
@@ -15,6 +15,7 @@ class SqsMessageProvider implements MessageProviderInterface
     protected $cache;
     protected $channel;
     protected $prefetch;
+    protected $waitTime;
     protected $queueName;
 
     /**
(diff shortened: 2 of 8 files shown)


composer.json
```diff
@@ -19,12 +19,15 @@
         "phpunit/phpunit": "~4.7|~5.0",
         "php-amqplib/php-amqplib": "~2.1",
         "doctrine/common": "~2.3",
-        "doctrine/dbal": "~2.2"
+        "doctrine/dbal": "~2.2",
+        "aws/aws-sdk-php": "2.*",
+        "phake/phake": "^2.3"
```
odolbeau: Is it possible to rewrite tests to use prophecy?

src/Swarrot/Broker/MessageProvider/SqsMessageProvider.php
```diff
@@ -0,0 +1,95 @@
...
+     * @param SqsClient                  $channel
+     * @param string                     $queueName
+     * @param MessageCacheInterface|null $cache
+     */
+    public function __construct(SqsClient $channel, $queueName, MessageCacheInterface $cache = null)
+    {
+        $this->channel = $channel;
+        $this->queueName = $queueName;
+        $this->cache = $cache?:new PrefetchMessageCache();
+        $this->prefetch = 9;
```
odolbeau: Could you add this as a constructor argument?

src/Swarrot/Broker/MessageProvider/SqsMessageProvider.php
```diff
@@ -0,0 +1,95 @@
...
+    public function get()
+    {
+        if ($message = $this->cache->pop($this->getQueueName())) {
+            return $message;
+        }
+
+        $result = $this->channel->receiveMessage([
+            'QueueUrl' => $this->getQueueName(),
+            'MaxNumberOfMessages' => $this->prefetch,
+            'WaitTimeSeconds' => 5,
```
odolbeau: Could you add this as a constructor argument?

src/Swarrot/Broker/MessageProvider/SqsMessageProvider.php
```diff
@@ -0,0 +1,95 @@
...
+    public function __construct(SqsClient $channel, $queueName, MessageCacheInterface $cache = null)
+    {
+        $this->channel = $channel;
+        $this->queueName = $queueName;
+        $this->cache = $cache?:new PrefetchMessageCache();
+        $this->prefetch = 9;
+    }
+
+    /**
+     * get.
```
odolbeau: inheritdoc (same below)

odolbeau:
Thanks! <3

Taluu:
The MessageCache classes are not within the scope of this PR IMO.

odolbeau:
@Taluu I think it is. RMQ also supports prefetch. The `MessageCacheInterface` could be used in several providers including the RMQ one.

Taluu:
Yes, but IMO, it should be in a different PR. I'm quite the partisan of "one pr does one thing" : here, it should just add the providers (and producers) for sqs. Another PR could add the prefetch. :}

odolbeau:
Not really a problem in this case. A dedicated PR for the `MessageCacheInterface` doesn't make sense as it's not usable without this provider.
IMO that's a prerequisite which is shipped in the same PR and it's fine like this. Furthermore, it's only 2 small classes. ;)

composer.json
```diff
@@ -16,15 +16,17 @@
         "psr/log": "~1.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "~4.7|~5.0",
+        "phpunit/phpunit": "~4.8.27|~5.0",
```
Taluu: why ? :o

stof: probbaly to make tests running with ``--prefer-lowest`` working when using the PHPUnit version installed through composer

composer.json
```diff
@@ -16,15 +16,17 @@
...
         "php-amqplib/php-amqplib": "~2.1",
         "doctrine/common": "~2.3",
-        "doctrine/dbal": "~2.2"
+        "doctrine/dbal": "~2.2",
+        "aws/aws-sdk-php": "~2.8"
```
stof: wouldn't it be better to build it against version 3 of the SDK ?

src/Swarrot/Broker/MessageProvider/SqsMessageProvider.php
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Class SqsMessageProvider.
+ */
+class SqsMessageProvider implements MessageProviderInterface
+{
+    protected $cache;
+    protected $channel;
+    protected $prefetch;
+    protected $queueName;
```
stof: @odolbeau I suggest using private properties. It makes it much easier to manage backward compatibility

src/Swarrot/Driver/MessageCacheInterface.php
```diff
@@ -0,0 +1,29 @@
...
+
+use Swarrot\Broker\Message;
+
+/**
+ * Interface MessageCacheInterface.
+ */
+interface MessageCacheInterface
+{
+    /**
+     * Pushes a $message to the end of the cache.
```
stof: I would say ``Pushes a message``. No need to use a variable name in the phpdoc

composer.json
nicolasThal: The sdk requires a version of php > 5.5 on version 3. 
https://github.com/aws/aws-sdk-php/blob/3.0.0/composer.json#L19

composer.json
nicolasThal: Yes, I had some issue with phpdocumentor.

https://github.com/phpDocumentor/ReflectionDocBlock/issues/80

nicolasThal:
@odolbeau I have updated the pr

composer.json
```diff
@@ -16,15 +16,17 @@
...
     },
     "suggest": {
         "pecl-amqp": "*",
         "php-amqplib/php-amqplib": "~2.1",
-        "symfony/console": "~2.4|~3.0"
+        "symfony/console": "~2.4|~3.0",
+        "aws/aws-sdk-php": ">=2.8"
```
odolbeau: I don't know this SDK at all.
Is it possible to have `^2.8 || ^3.0` maybe?

odolbeau:
Looks good right now. :)
Can you rebase your branch and change the aws-sdk dependency please? :)

nicolasThal:
@odolbeau i have rebase the branch.

For the sdk version, I perfer to use the version 2, the 3 has a lot of bc breaks.

Taluu:
with your `">= 2.8"` constraint, it * will * install the 3.0 if it is available. So you should at least change this to `^2.8`, (which means `>=2.8, < 3.0`). But you should also consider of trying to make something compatible with both `2.8` and `3.0` imo

nicolasThal:
@Taluu I have already put `^2.8` for the version.

As the method to get the element from the sqs queue are not the same in the sdk, I will be a lot of work to make the code compatible with both sdk version

odolbeau:
Looks good for me!
Is it ok for you @Taluu & @stof?

odolbeau:
Looks like it's OK for everyone.

Again, thanks for this awesome contribution @nicolasThal! :)

PR #135, opened 2016-09-28
Add more keywords for better discoverability on Packagist

stof:


stof:
Regarding the test failure, it looks like a conflict between the dependencies installed locally and the one available in the PHPUnit 4.8.27 phar (which was not forcing to eager-load all its classes yet, as done in recent PHPUnit versions IIRC).
there is 2 ways to fix it:
- use the composer-managed PHPUnit version on Travis (it is installed anyway, so there is no additional cost)
- remove the composer-managed PHPUnit version

odolbeau:
Correction merged, you can rebase your PR. :)

stof:
well, do you actually need a rebase to see that it does not break anything ? This PR does not change any code (and we still have several passing jobs in the existing build) :smile:

odolbeau:
Good point. :p

PR #136, opened 2016-09-28
Use composer-managed phpunit version

odolbeau:
Linked to #135


stof:
you have ``bin-dir: bin`` in your composer.json, so the path to PHPUnit is wrong (I would suggest removing the bin-dir overwrite though)

PR #138, opened 2016-12-01
Log the exception when stopping the retry as well

stof:
When retrying, the exception being catched was logged, but not when stopping the retry


odolbeau:
Thanks!

PR #139, opened 2016-12-12
Allow custom log level by thrown exception

yguedidi:
This is just a POC.

Missing things:
- [x] Update `InstantRetryProcessor`
- [x] Documentation
- [x] Tests

Usage:
Sometimes, you want to retry a message because it's just not yet ready to be processed (for exemple due to an extrenal cause/provider/...), but you don't want to log those retry as warning to not flood your warning logs with those irrelevant logs.

```php
class NotReadyException extends \Exception
{
}
```
```php
    ->push(
        'Swarrot\Processor\Retry\RetryProcessor',
        [
            'retry_key_pattern' => 'foo',
            'retry_log_levels_map' => [NotReadyException::class => LogLevel::INFO],
        ]
    )
;
```

Then in the processor:
```php
if (!$object->isReadyToProcess()) {
    throw new NotReadyException('Retry later, the object is not yet ready.');
}
```

What do you think about this?

### DIFF
--- a/src/Swarrot/Processor/InstantRetry/InstantRetryProcessor.php
+++ b/src/Swarrot/Processor/InstantRetry/InstantRetryProcessor.php
@@ -2,6 +2,7 @@
 
 namespace Swarrot\Processor\InstantRetry;
 
+use Psr\Log\LogLevel;
 use Swarrot\Broker\Message;
 use Swarrot\Processor\ConfigurableInterface;
 use Swarrot\Processor\ProcessorInterface;
@@ -54,6 +55,7 @@ class InstantRetryProcessor implements ConfigurableInterface
         $resolver->setDefaults(array(
             'instant_retry_delay' => 2000000,
             'instant_retry_attempts' => 3,
+            'instant_retry_log_levels_map' => array(),
         ));
     }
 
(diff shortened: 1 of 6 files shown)


odolbeau:
I understand the problem and your PR looks good right now. :)
I'm not a big fan of adding the `NotReadyException` directly inside swarrot for 2 reasons:
* there could be several reason to not log in `warning`
* you'll have to change your business exception to makes them extends the swarrot one

yguedidi:
@odolbeau thanks for you're reactivity!
I'm not suggesting to add a `NotReadyException` inside swarrot. It is a business exception, created by the user, that's why the map is controller by the user.
It may be named `SleepAndWakeLaterException` or be an interface :)

odolbeau:
Oh yeah! I thought you would like to add the `isReadyToProcess` call inside the `RetryProcessor` but you were talking about your own processor, makes sense!
Some mornings are harder than others. ;)

yguedidi:
Haha I understand! So would you accept this PR if I finish it? :)

odolbeau:
Sure! :)

yguedidi:
Thanks @odolbeau! I'll work on it went I find some free time :) (ASAP I hope)

yguedidi:
@odolbeau updated! I think this is ready :)

odolbeau:
Yes it is! Sorry for the delay and thanks a lot for this. :)

yguedidi:
Thanks @odolbeau!! When will you tag a 2.3.0 release please? :)

odolbeau:
Soon! ;) #141

PR #143, opened 2016-12-29
Handle RabbitMQ case when using amqp php extension

olaurendeau:
With amqp c extension `headers` are in a separate key, so I propose to array_merge it with properties in order to have access to `correlation_id` & `reply_to` keys

### DIFF
--- a/src/Swarrot/Processor/RPC/RpcServerProcessor.php
+++ b/src/Swarrot/Processor/RPC/RpcServerProcessor.php
@@ -35,7 +35,7 @@ class RpcServerProcessor implements ProcessorInterface
     {
         $result = $this->processor->process($message, $options);
 
-        $properties = $message->getProperties();
+        $properties = $this->getProperties($message);
 
         if (!isset($properties['reply_to'], $properties['correlation_id']) || empty($properties['reply_to']) || empty($properties['correlation_id'])) {
             return $result;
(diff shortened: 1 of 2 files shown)


odolbeau:
Shouldn't it be done inside the MessageProvider?
Look here: https://github.com/swarrot/swarrot/blob/master/src/Swarrot/Broker/MessageProvider/PeclPackageMessageProvider.php#L49-L50
Maybe it could be a better solution to test the content of headers & the return value of `getCorrelationId` & `getReplyTo` to choose the right one?
This way if another processor need those informations they will be available in properties.

stof:
I agree with @odolbeau. Differences between the different providers must be handled in providers.

olaurendeau:
Totally make sense :)

Digging a bit more I realize that everything is fine, I have been mislead by someone arguing that `reply_to` & `correlation_id` are removed from the basic property set.

So I was doing
```php
	$message = new Message(
		'payload',
		['headers' => [ 'reply_to' => 'queue.response', 'correlation_id' => '12EDAZD12D']]
	);
```
while
```php
	$message = new Message(
		'payload',
		['reply_to' => 'queue.response', 'correlation_id' => '12EDAZD12D']
	);
```
perfectly worked ...

Sorry for your time and happy new year !

PR #147, opened 2017-03-07
Add support for publisher confirms

borisbabic:
This is adds support for publisher confirms (see [here](https://www.rabbitmq.com/confirms.html)) while retaining backwards compatibility with custom publishers.

If this is accepted I will also add support for publisher confirms in the symfony bundle.


src/Swarrot/Broker/MessagePublisher/PublishConfirmPublisherInterface.php
```diff
@@ -0,0 +1,24 @@
+<?php
+declare(strict_types=1);
```
stof: should be removed. We need to be compatible with PHP 5 (and btw, this is totally useless for an interface, as it is not doing any method call)

src/Swarrot/Broker/MessagePublisher/PublishConfirmPublisherInterface.php
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Swarrot\Broker\MessagePublisher;
+
+/**
+ * Interface PublishConfirmPublisherInterface
```
stof: this phpdoc is totally useless. It should explain what the interface represent, not what its name is (I can already know it by reading the next line)

stof:
I'm quite -1 for the new interface. I don't see any reason to implement this feature as a mutable state. This should use configuration through the constructor (which removes any question about support in other publishers btw, as constructors are free to accept the configuration settings they want).

stof:
@odolbeau I suggest using private visibility for any new property, to make BC easier. Protected APIs are always the worse ones regarding BC layers (depending on what they are about, they may make a BC layer totally impossible to build, especially when dealing with properties rather than methods)

borisbabic:
How "connected" are the bundle and this library (basically, is it okay for the bundle to influence implementations here)?  The bundle influenced my implementation as it does not allow much flexibility when it comes to publishing.

stof:
Well, the bundle can be updated to support new configuration settings when we introduce them.

Switching to confirm mode should be part of the definition of a publisher IMO. Using a mutator allowing calling code to turn it on is a bad idea, especially given that it cannot be turned off again, as this will impact other code using the same publisher.

borisbabic:
You're right, thanks. I'll change the implementation.

odolbeau:
There nothing to add here, thanks @stof!

I agree on properties visibility: using `private` when possible is definitively a good habit.
I'll try to keep this in mind in order to change properties visibility before releasing the next major version.

src/Swarrot/Broker/MessagePublisher/PeclPackageMessagePublisher.php
```diff
@@ -19,6 +25,54 @@ public function __construct(\AMQPExchange $exchange, $flags = AMQP_NOPARAM, Logg
...
+        }
+        return $this->nackHandler;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function enterConfirmMode($timeout = 0)
+    {
+        if (!(method_exists($this->exchange->getChannel(), "confirmSelect"))) {
```
Taluu: why the `!(...)` instead of `!...` ?

borisbabic:
No good reason, I'll remove it.

src/Swarrot/Broker/MessagePublisher/PeclPackageMessagePublisher.php
borisbabic: No good reason, I'll delete it

odolbeau:
Looks good for me.
You still have some CS errors, could you please run php-cs-fixer?

borisbabic:
I fixed the php-cs issue (sorry, should have done that before updating the pull request). 

I'll add a publisher confirm test for both publishers.

borisbabic:
I added a test for each publisher. 

While doing so I noticed stylistic differences between the two test classes, specifically when it comes to chained method calls. I couldn't find some mention of a preferred style for this project so I kept my code consistent with each class. 

If there is a preferred style I can make the test classes consistent with each other.

src/Swarrot/Broker/MessagePublisher/PeclPackageMessagePublisher.php
```diff
@@ -11,14 +11,59 @@ class PeclPackageMessagePublisher implements MessagePublisherInterface
...
+                    unset($this->pendingMessages[$tag]);
+                }
+            } else {
+                unset($this->pendingMessages[$deliveryTag]);
+            }
+
+            if (count($this->pendingMessages) > 0) {
+                return true; //still need to wait
+            }
+            return false;
```
odolbeau: missing empty line before return statement

odolbeau:
There isn't any preferred style. :)
There is 1 missing empty line and this PR is ready to be merged. :+1:

src/Swarrot/Broker/MessagePublisher/PeclPackageMessagePublisher.php
borisbabic: Before `return false;`? My versions of php-cs-fixer (2.0-DEV) and phpcs (2.6.0) don't find anything, probably because it is an anonymous function.

odolbeau:
Thanks for your contribution. :+1: 
I'll try to release a new version this week.

PR #149, opened 2017-03-21
Doc improvements

greg0ire:


README.md
```diff
@@ -45,9 +45,13 @@ from the broker. This processor must implement
...
 
-class Processor implements ProcessorInterface {
-    public function process(Message $message, array $options) {
+class Processor implements ProcessorInterface
+{
+    public function process(Message $message, array $options)
+    {
         echo sprintf("Consume message #%d\n", $message->getId());
+
+        return true; // Continue processing other messages
```
greg0ire: On top of documenting the return value, this might be a bit more future proof, if one day this lib requires php7 and you decide to enforce boolean as type hint for the return value.

odolbeau: Right now, a lot of processors (both in swarrot and in proprietary software) don't return a boolean (a lot of them return null).
Even if I'm not sure it's a good idea to enforce boolean as type hint for the return value (at least in the near future) it's still a good idea to add this in the documentation. :)
Maybe it will ease the migration for some people when we'll decide to add this type hint.

odolbeau:
:+1:

PR #153, opened 2017-06-20
PHP7 1 now is a requirement

### LINKED ISSUE #152: Bump php minimal version
The php 5.4 version is not supported anymore and php 5.6 only receive security patch.

Shouldn't we bump the minimal php version to php7.0?

odolbeau:
Fix #152

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,13 @@
 language: php
 
-php: [5.4, 5.5, 5.6, 7.0, hhvm]
-
 matrix:
     include:
+        # Test with default configuration
+        - { php: 7.1 }
         # Test against librabbitmq cutting-edge
-        - { php: 5.5, env: LIBRABBITMQ_VERSION=master }
-        - { php: 5.4, env: COMPOSER_FLAGS='--prefer-lowest --prefer-stable' }
-    allow_failures:
-        - php: hhvm
+        - { php: 7.1, env: LIBRABBITMQ_VERSION=master }
+        # Test with lowest dependencies
+        - { php: 7.1, env: COMPOSER_FLAGS='--prefer-lowest --prefer-stable' }
 
 services: [rabbitmq]
 
--- a/composer.json
+++ b/composer.json
@@ -11,12 +11,12 @@
         }
     ],
     "require": {
-        "php": "^5.4 || ^7.0",
-        "symfony/options-resolver": "^2.3 || ^3.0",
+        "php": "^7.1",
+        "symfony/options-resolver": "^2.8 || ^3.0",
         "psr/log": "^1.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "^4.8.27 || ^5.0",
+        "phpunit/phpunit": "^6.2",
         "php-amqplib/php-amqplib": "^2.1",
         "doctrine/common": "^2.3",
         "doctrine/dbal": "^2.2",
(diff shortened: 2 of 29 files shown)


.travis.yml
```diff
@@ -1,14 +1,15 @@
 language: php
 
-php: [5.4, 5.5, 5.6, 7.0, hhvm]
+php: [7.1]
 
 matrix:
     include:
+        # Test with default configuration
+        - { php: 7.1 }
```
stof: this is duplicating the job added though `php: [7.1]`

composer.json
```diff
@@ -11,12 +11,12 @@
         }
     ],
     "require": {
-        "php": "^5.4 || ^7.0",
-        "symfony/options-resolver": "^2.3 || ^3.0",
+        "php": "^7.1",
+        "symfony/options-resolver": "^3.0",
```
stof: Can we use `^2.8 || ^3.0` ? Otherwise, people using the Symfony LTS will be blocked on older versions. 2.8 contains all the features we have in 3.0

src/Swarrot/Processor/MaxMessages/MaxMessagesProcessor.php
```diff
@@ -61,17 +61,13 @@ public function process(Message $message, array $options)
...
 
-        if (method_exists($resolver, 'setDefined')) {
-            $resolver->setAllowedTypes('max_messages', 'int');
-        } else {
-            // BC for OptionsResolver < 2.6
-            $resolver->setAllowedTypes(array(
-                'max_messages' => 'int',
-            ));
-        }
+        $resolver;
```
stof: dead code

.travis.yml
odolbeau: Without this I had only 2 jobs. See: https://travis-ci.org/swarrot/swarrot/builds/244940684

It looks OK right now: https://travis-ci.org/swarrot/swarrot/builds/244942392

Any idea of what's wrong in the config?

odolbeau:
@stof should be OK now. :)

olaurendeau:
@stof Do you approve ? :)

olaurendeau:
@odolbeau To me it should be a new major version v3.0.0, what do you think ?

odolbeau:
Yes. Thé composer.json file is up to date, you just have to tag a new version. :)

olaurendeau:
Done https://github.com/swarrot/swarrot/releases/tag/v3.0.0 :)

Taluu:
IMO it's not worth bumping a major release, as nothing has really changed in the API. Just a bump in the dependency, and a minor is usually enough (composer handles this quite nicely actually).

http://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api

> ### What should I do if I update my own dependencies without changing the public API?
> 
> That would be considered compatible since it does not affect the public API. Software that explicitly depends on the same dependencies as your package should have their own dependency specifications and the author will notice any conflicts. Determining whether the change is a patch level or minor level modification depends on whether you updated your dependencies in order to fix a bug or introduce new functionality. I would usually expect additional code for the latter instance, in which case it’s obviously a minor level increment.

Just saying.

olaurendeau:
That's a good point @Taluu :)
To me, releasing a major version keep us opportunity to patch on top of v2.4.0 if needed

Taluu:
If you need to patch on `v2.4.0`, make a `v2.4.1` tag :P

olaurendeau:
Oh sorry you mean that `v2.5.0` would have been enough :) It's not too late, @odolbeau what do you think ?

odolbeau:
Not sure. It's not only the dependencies which have been updated but also the requirements... :/

Taluu:
Which composer handles quite nicely, again. No changes in the public api.

PR #155, opened 2017-07-19
Queue interop producer\provider

makasim:
The PR adds provider\publisher that can work with any [queue interop](https://github.com/queue-interop/queue-interop) compatible transport.

TODO:

* [x] add tests
* [x] add doc

I welcome everyone who is interested in MQs to join the queue interop group


olaurendeau:
Hello @makasim, queue-interop look like a great idea !
capture d ecran 2017-07-19 a 15 11 00
Looking at compatible queue systems list (and not being an expert at all), I'm just wondering : 
AMQP use the notion of routing_key, notion that you didn't implement in your producer because obviously it's not an avery queue system feature. Some Processors will not be compatible with queue-interop for example the retry one https://github.com/swarrot/swarrot/tree/master/src/Swarrot/Processor/Retry

As swarrot is theoretically built to be queue agnostic, but currently only support AMQP, do we need a mechanism explaining / constraining those limitations if swarrot have ambition to support other queueing systems ?

makasim:
The AMQP implementation  [supports the ability to publish messages on the exchange with routing key set](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/amqp-ext/AmqpProducer.php#L57). It has to be set on the topic object with `setRoutingKey` method. But the method is not a part of the speck (neither queue-interop nor origin JMS). 

If the `$key` is used to publish the message to a queue with the same then it could be implemented. The amqp producer can publish messages on a queue without the need to have explicit exchange. 

```php
<?php
    public function publish(Message $message, $key = null)
    {
        // ....

        $this->producer->send(
            $key ? $this->context->createQueue($key) : $this->topic,
            $this->context->createMessage($message->getBody(), $properties, $headers)
        );
    }
```

olaurendeau:
> The amqp producer can publish messages on a queue without the need to have explicit exchange.
Indeed and in this case routing_key will be used to define the destination queue name. 

```
$this->producer->send(
            $key ? $this->context->createQueue($key) : $this->topic,
            $this->context->createMessage($message->getBody(), $properties, $headers)
        );
```
If I'm not mistaken this would use the routing_key to address the queue having the name of the routing_key value ? So this is not supporting AMQP bindings.

For RPC it makes perfect sense as discussed here https://github.com/swarrot/SwarrotBundle/pull/94

But in the example of the Retry processor a message is publish on an exchange and the routing_key is used to dispatch the message on the proper queue through the AMQP bindings. So this processor would not be compatible with Queue interop, which would maybe be fine if we had a mechanism constraining processors to a set of Publishers / Providers, meanwhile I assume that specifying it in documentation would be enough 😃

olaurendeau:
Yes this is precisely what I wanted to emphasis by the example of https://github.com/swarrot/SwarrotBundle/pull/94 :)

For the RetryProcessor the `amp.direct` + queue name as routing key feature will not do the trick.
Routing keys will be a pattern like `retry_%attempt` and the resulting key `retry_10`, `retry_9`, ... , `retry_1` that's why a binding `retry_#` between the retry exchange and the destination queue is necessary.

Anyway my point is simply that the Queue Interop Publisher / Provider documentation should precise that RetryProcessor will not be compatible :)

makasim:
@olaurendeau finally got it. I'll mention it in the docs

makasim:
btw we are thinking of building an amqp-interop on top of queue-interop. That would make it possible to support such AMQP specific cases without coupling to implementation.

olaurendeau:
Make sense :) And there is others protocols providing specific features, like gearman.
Do you plan to move php-enqueue/* repos to queue-interop ?

makasim:
> Do you plan to move php-enqueue/* repos to queue-interop ?

I dont think it worth it.

makasim:
I am finished. Could you please have a look?

src/Swarrot/Processor/InteropProxyProcessor.php
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function process(PsrMessage $message, PsrContext $context)
+    {
+        $properties = $message->getHeaders();
+        $properties['headers'] = $message->getProperties();
+
+        $this->swarrotProcessor->process(new Message($message->getBody(), $properties), $this->options);
```
olaurendeau: Why is this proxy useful as `InteropMessageProvider::get` will return a `Swarrot\Broker\Message` ?

tests/Swarrot/Processor/InteropProxyProcessorTest.php
```diff
@@ -0,0 +1,207 @@
...
+    public function testInstance()
+    {
+        $swarrotProcessor = $this->prophesize('Swarrot\Processor\ProcessorInterface');
+        $this->assertInstanceOf(
+            'Interop\Queue\PsrProcessor',
+            new InteropProxyProcessor($swarrotProcessor->reveal())
+        );
+    }
+
+    public function test_get_with_messages_in_queue_return_message()
```
olaurendeau: Not sure about test name :)

tests/Swarrot/Processor/InteropProxyProcessorTest.php
```diff
@@ -0,0 +1,207 @@
...
+
+        $processor = new InteropProxyProcessor($swarrotProcessor->reveal(), ['fooOpt' => 'fooOptVal']);
+
+        $result = $processor->process($message->reveal(), $context->reveal());
+
+        $this->assertSame(PsrProcessor::ACK, $result);
+
+
+    }
+//
```
olaurendeau: Maybe remove those commented lines ?

README.md
```diff
@@ -105,6 +105,49 @@ Here is an illustration to show you what happens when this order is used:
...
+use Swarrot\Broker\Message;
+
+/** @var \Interop\Queue\PsrContext $context */
+
+$messagePublisher = new InteropMessagePublisher($context, 'aTopic');
+
+$messagePublisher->publish(new Message('aBody'));
+```
+
+_**Note: ** The second argument `key` is not supported at the moment._
```
olaurendeau: Maybe the place to also inform people that RetryProcessor is not supported by queue-interop publisher ?

src/Swarrot/Processor/InteropProxyProcessor.php
makasim: @olaurendeau sorry, I dont get it.

odolbeau:
Hello @makasim.
Thanks for this PR. I won't be able to review it correctly before 2-3 weeks as I don't have my computer with me at the moment.
Anyway it looks good for me and it's definitively a good thing to be compatible with queue-interop!
@olaurendeau feel free to merge my PR for PHP 7.1. You should be able to tag a new version. :) BTW, don't wait for me to merge this PR as soon as you think it's ready.

I'll take some time to dig into queue-interop group / repositories ASAP.

src/Swarrot/Processor/InteropProxyProcessor.php
olaurendeau: I don't see the point of the InteropProxyProcessor :)

olaurendeau:
@makasim Instead of changing .travis.yml could you simply rebase from master ? :)
Thanks !

src/Swarrot/Processor/InteropProxyProcessor.php
makasim: The InteropProxyProcessor is useful outside Swarrot, but still using its processors. 
For example you can use it with [enqueue consumption](https://github.com/php-enqueue/enqueue-dev/blob/master/docs/quick_tour.md#consumption) including cli commands or [enqueue simple client](https://github.com/php-enqueue/enqueue-dev/blob/master/docs/quick_tour.md#client)

olaurendeau: Hum, I'm more confident about the Message Provider & Publisher, may the ProxyProcessor can be the object of another PR, so that people can discuss about it ?

makasim: as you wish, gonna remove it

makasim:
@olaurendeau done

makasim:
That would be possible to support routing key once we finalize amqp interop https://github.com/queue-interop/amqp-interop

makasim:
Don't you mind I add Swarrot to the projects list on queue-interop repo? If someone wants to participate I am more than happy to send you an invite. 

Is there any ETA on when this PR is merged?

olaurendeau:
Hello @makasim,
I would have liked another review maybe from @stof or @Taluu 
Let's say that next monday if no other review, I'll merge and release :)

Yes feel free to add swarrot to queue-interop project list :)

makasim:
@olaurendeau thanks for the update. Monday is fine for me.

makasim:
add key arg support, it will work for all amqp interop based transports

olaurendeau:
As soon as travis is happy I'll merge it 👍 
Thanks @makasim !

makasim:
@olaurendeau it's happy

olaurendeau:
It's released @makasim https://github.com/swarrot/swarrot/releases
Thanks a lot for your contribution :)

makasim:
@olaurendeau Would you be interested to join queue interop? I'll be more than happy to send you an invite.

olaurendeau:
Well it would be with pleasure, but I'm not sure to be able to contribute actively, having quite a lot of project running lately :)

makasim:
> but I'm not sure to be able to contribute actively, having quite a lot of project running lately :)

Don't worry.

PR #160, opened 2017-11-01
Update rabbitmq-c installation script

odolbeau:
Support of autotools have been removed from https://github.com/alanxz/rabbitmq-c

We have to use `clang` instead.


odolbeau:
Same PR on php-amqp: https://github.com/pdezwart/php-amqp/pull/295

PR #162, opened 2017-11-11
Use ::class syntax in tests

notFloran:
The latest tests use the syntax ":: class" but the old still use the class name as a string.

To be consistent, I modified all the tests to use the ":: class" syntax


odolbeau:
Thanks a lot. :)

PR #167, opened 2018-01-07
[enhancement] Replace Message with MessageInterface

Swahjak:
| Q             | A
| ------------- | ---
| Branch?       | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Currently we're restricted to using Swarrot\Broker\Message as our Message format. I would like to introduce Swarrot\Broker\MessageInterface to allow custom implementations of messages (and providers / publishers). Also added some basic functionality (setters, getProperty, setProperty) to the MessageInterface.

Updated docs (including some old references to MessageProviders).

Used https://github.com/queue-interop/queue-interop/blob/master/src/PsrMessage.php for inspiration.


odolbeau:
Hi @Swahjak and thanks for your contribution!

In the current implementation, a `Message` is  immutable. With setters in the interface it's not the case anymore and I'm not sure I like this idea.
I used this library a lot (obviously) and I never had to update a Message. IMO, as soon as you need to update an existing message, you actually need to create another one.

Do you have a concrete use case regarding message update?

Regarding the interface itself (and if removing the immutable state of a Message makes sense), do you think it's a good idea to use directly the `PsrMessage` interface from queue-interop?

@makasim Sorry to bother you. :/ What do you think about this? Could you explain why you chose to have a non-immutable Message?

Swahjak:
Ok, I can see the point regarding the Immutability but it does add a convenience in my opinion. Why recreate a message (in a retry) when I already have the object. 

An example for for the case would be an SQS messages. The API allows for more message specific parameters than the current implementation allow (without 'abusing' properties). DelaySeconds, MessageDeduplicationId https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessage.html. 

To be more specific, I would like to introduce (in my project) as processor that allows me to delay messages based on the retry attempts. The first retry could be in 5 seconds, but maybe the second one would be 5 minutes (just as an example). Using an interface I could detect a specific message class (that doesn't necessarily extend Swarrot\Broker\Message) and apply my custom logic. 

As to using the ```PsrMessage```; I'm not sure whether the ```PsrMessage``` is what a common message should look like. Simply haven't used enough queuing frameworks to give an opinion about that.

Taluu:
I'm not convinced addind an interface is actually relevant. The `Message` is a value-object that will be rebuilt when receiving a message anyway,

If you want to make it mutable, you could always extend it and add the setter (the properties are protected).

So 👎 for me.

About using the `PsrMessage`, why not though. But as it's not a real PSR...

Swahjak:
@Taluu like I said ```But if immutability makes more sense I would be happy to make the interface immutable.```. That still leaves my case as to why an Interface *could* make sense.

Taluu:
It's not really the immutable point that bothers me (well, it does a little though). It's the fact that trying to make an interface out of a value object (because that's what `Message` is). Even though the `PsrMessage` is a mutable interface, it would still be converging with a standard of some sort.

But as I said, it's not really a PSR (as there is no psr actually), so I'm actually not really in favor or against that. I'm just against making an interface out of the `Message` value object, I'm failing to see the point.

makasim:
queue interop is inspired by [Java JMS] standard (matured one, de jure and de facto standard in Java\Enterpise world). Their message is [mutable](https://docs.oracle.com/javaee/7/api/javax/jms/Message.html). 

I haven't experienced any problems because of mutability so far. 

> But as I said, it's not really a PSR (as there is no psr actually), so I'm actually not really in favor or against that.

I've actually tried stuck up a [conversation](https://groups.google.com/forum/#!msg/php-fig/CcFN7zr90Wo/3ihpslRDBwAJ) on PHP FIG mailing list and tried to reach out some guys from the group.  Without success though. So I decided to go a different way, that's how queue interop appeared.

metfan:
@Swahjak about your idea of change ttl on different retry. I did it on previous project. I create a specific processor for that based on specific  messages properties I added to my message. This header was just a counter and ttl for retry was based on the value of my counter. You don't need more than what Swarrot already propose. But yes because of immutability you have to create another message to upgrade the counter on the message header.

stof:
Immutability of the value object also simplifies some code. For instance, if you want to re-dispatch a message with some change, mutating the current value object has a bad effect: it also affect the current message, not only the re-dispatched one, which will be observable by processors in the stack when going out. The immutability avoids lots of complexity here actually.

Ngob:
Hello,

I am convinced about the usage of a MessageInterface, it would help me to force the user of my code to push a message with correct format. It help me to add some basic field to my events (as the date and the current node pushing the message, it help a lot when debugging)

At the moment, I am stuck with message factories

odolbeau:
Hi @Ngob 

As explained in this issue, the `Message` class is immutable on purpose. Having a interface will break this immutability and could introduce some unwanted behaviors in existing applications. Having messages factories isn't a problem IMO.

In your case, I don't understand how a `MessageInterface` will change anything? :thinking:
 
Why an interface containing the exact same methods will help you achieve your needs and why is the actual `Message` class a problem for your use case?

Ngob:
Hello @odolbeau, Thanks for your answer!

I am not sure I understand the reason why an interface would break immutability (I mean, there are interfaces that does not break immutability, for example the `MessageInterface` from PSR7). Of course, I agree with you that there should not be "setter" inside the interface.
An interface would help, I would be able to write my own  `message`. Instead to construct the object using a string, I want to construct an object using an JsonSerializable object, and ensure that all my message are Json compilant and have the exact same format (with additional fields like the current date).
So, in a microservice, I can share the "message" class and reuse it. This way I am ensuring that all my message have all minimum required field.
For the moment I am using a factory pattern to construct a SwarrotMessage from a JsonSerializable. I am typehinting on a swarrotmessage, so it feel like I am dependent of the message implementation.

In my opinion, all class should be either abstract or final and I should not be dependent of a particular implementation of a message, but on a public API. But for this, I need interfaces

stof:
To me, this looks like you should be building your own abstraction *around* swarrot rather than expecting to build custom implementations of the swarrot message.
Btw, if the suggested MessageInterface is used in all typehints, you would still have nothing enforcing that *your* implementation of the interface is being used.

Issue #106, opened 2015-07-31
Use a consume instead of a get in the Consumer


Taluu:
Why on the Consumer it is always a `get` instead of a `consume`, which ideally does blocked message connections ? 

An idea would be to add a new `consume` method in the `MessageProviderInterface` (3.0 ? or use another interface which specifies a `consume` method, to stay on 2.X ? cc @stof), and implement it in the different native provider (Pecl, Oldsound) and an option in the Consumer to choose between using an infinite loop or the consume mechanism instead (which is the current mechanism). Or offer up a new Consumer which does about the same thing as the current one, but uses `consume` instead, instead of an option.

WDYT ? I'd be glad to try a POC on that. cc @krichprollsch

odolbeau:
The `consume` method is a blocking call.
It means you can't use all processor which implements the `SleepyInterface`. No more `MaxMessagesProcessor`, `MaxExecutionTimeProcessor` and even worse, no more `SignalHandlerProcessor`.

What's your use case?
Why do you want to use the `consume` method instead of the `get` one?

stof:
Blocking until there is a message is less flexible than waiting externally, as it removes control on whether we want to wait. for instance, implementing the InsomniacProcessor is impossible if you take back control over waiting to the provider.

stof:
and this would also make it harder to implement providers for other systems. For instance, if you build a Redis-backed or a MongoDB-backed provider (Resque-like), you would have to use an infinite loop inside the provider to wait until there are messages

Taluu:
Yep, that's why I suggested, instead of having an option or modify the current Consumer, to another consumer object (`BlockingConsumer` ?), with another interface which tells a provider to be able to use a `consume` method on it. This way, BC would be provided, and other providers (Such as the Resque-like ones you mentionned @stof) would still be able to work with the classic Consumer ?

stof:
@Taluu but what would be the benefit of using a ``BlockingConsumer`` ? AFAICT, you are only *loosing* important Swarrot features by using it instead of the current consumer.
Not being able to stop your worker at all because the SignalHandlerProcessor does not work anymore seems a bad idea.

NoiseByNorthwest:
@stof this is not true, signal handling responsiveness is achievable simply by setting a finite connection timeout as described here: http://stackoverflow.com/questions/15487657/rabbitmq-amqpconsume

[...]

PR #169, opened 2018-01-10
[DRAFT][Need feedbacks] Consume POC

olaurendeau:
## Context

In Go or Javascript I've notice that you can see how many consumers are connected, what are their names and with which arguments they have been popped

capture d ecran 2018-01-09 a 15 20 38

After searching a bit I understood that we can also achieve it in PHP and with swarrot, by using the consume method instead of the get method.

## Proposition

So my suggestion would be to add the possibility to choose either the consume or the get method when consuming.

Attached code is just an ugly draft to test it. I'd like to know if someone see an issue with this idea before going further ?

|  | Get | Consume |
| ---  | --- | --- |
| Benefits | Non-Blocking | Long polling, Faster, show consumer in RMQ UI |
| Drawbacks | Short polling, Treatments in SleepyInterface load the host | Blocking |

Exemple

capture d ecran 2018-01-10 a 21 19 43

## Measures

Blackfire profiles
- Get : https://blackfire.io/profiles/d63a7eef-fe40-476b-8850-d190334b6362/graph
- Consume  : https://blackfire.io/profiles/dd9bbb8f-f7e5-4aa7-83d9-bd036b74793d/graph

Consume seems slightly more performant (~10%) than get
```
vagrant@local:~/www/lafourchette-b2c-crm(master)$ time app/console swarrot:consume:customer_consumer --max-messages=1000 --method=get

real	0m7.405s
user	0m2.476s
sys	0m0.228s
vagrant@local:~/www/lafourchette-b2c-crm(master)$ time app/console swarrot:consume:customer_consumer --max-messages=1000 --method=consume

real	0m5.400s
user	0m1.952s
sys	0m0.172s
vagrant@local:~/www/lafourchette-b2c-crm(master)$ time app/console swarrot:consume:customer_consumer --max-messages=10000 --method=get

real	1m0.685s
user	0m20.997s
sys	0m1.448s
vagrant@local:~/www/lafourchette-b2c-crm(master)$ time app/console swarrot:consume:customer_consumer --max-messages=10000 --method=consume

real	0m53.869s
user	0m19.749s
sys	0m1.196s
```

### DIFF
--- a/src/Swarrot/Broker/MessageProvider/MessageProviderInterface.php
+++ b/src/Swarrot/Broker/MessageProvider/MessageProviderInterface.php
@@ -14,6 +14,14 @@ interface MessageProviderInterface
     public function get();
 
     /**
+     * consume
+     *
+     * @param string $consumerTag Identify the consumer for the rabbitmq-server
+     * @param callable $callback A callback receiving a Swarrot\Broker\Message instance on consumption
+     */
+    public function consume($consumerTag, callable $callback);
+
+    /**
      * ack.
      *
      * @param Message $message
--- a/src/Swarrot/Consumer.php
+++ b/src/Swarrot/Consumer.php
@@ -63,6 +63,7 @@ class Consumer
         $this->optionsResolver->setDefaults(array(
             'poll_interval' => 50000,
             'queue' => $this->messageProvider->getQueueName(),
+            'method' => 'consume'
         ));
 
         if ($this->processor instanceof ConfigurableInterface) {
@@ -75,24 +76,53 @@ class Consumer
             $this->processor->initialize($options);
         }
 
+        if ($options['method'] === 'consume') {
+            $this->methodConsume($options);
+        } else {
+            $this->methodGet($options);
+        }
+
+
+        if ($this->processor instanceof TerminableInterface) {
+            $this->processor->terminate($options);
+        }
+    }
+
+    private function methodConsume(array $options = array())
+    {
+        $consumerTag = uniqid();
+        try {
+
+            $this->messageProvider->consume($consumerTag, function($message) use ($options) {
+
+                if (false === $this->processor->process($message, $options)) {
+                    throw new \Exception('Stop processing');
+                }
+
+                if ($this->processor instanceof SleepyInterface) {
+                    if (false === $this->processor->sleep($options)) {
+                        throw new \Exception('Stop processing');
+                    }
+                }
+            });
+
+        } catch (\Exception $e) {}
+    }
+
+    private function methodGet(array $options = array())
+    {
         while (true) {
             while (null !== $message = $this->messageProvider->get()) {
                 if (false === $this->processor->process($message, $options)) {
                     break 2;
                 }
             }
-
             if ($this->processor instanceof SleepyInterface) {
                 if (false === $this->processor->sleep($options)) {
                     break;
                 }
             }
-
             usleep($options['poll_interval']);
-        }
-
-        if ($this->processor instanceof TerminableInterface) {
-            $this->processor->terminate($options);
         }
     }
 
--- a/src/Swarrot/Broker/MessageProvider/PeclPackageMessageProvider.php
+++ b/src/Swarrot/Broker/MessageProvider/PeclPackageMessageProvider.php
@@ -22,6 +22,48 @@ class PeclPackageMessageProvider implements MessageProviderInterface
     /**
      * {@inheritdoc}
      */
+    public function consume($consumerTag, callable $callback)
+    {
+
+        $this->queue->consume(function(\AMQPEnvelope $envelope) use ($callback) {
+
+            if (!$envelope) {
+                throw new \Exception('No envelope');
+            }
+
+            return !$callback(new Message(
+                $envelope->getBody(),
+                array(
+                    'content_type' => $envelope->getContentType(),
+                    'routing_key' => $envelope->getRoutingKey(),
+                    'delivery_tag' => $envelope->getDeliveryTag(),
+                    'delivery_mode' => $envelope->getDeliveryMode(),
+                    'exchange_name' => $envelope->getExchangeName(),
+                    'is_redelivery' => $envelope->isRedelivery(),
+                    'content_encoding' => $envelope->getContentEncoding(),
+                    'type' => $envelope->getType(),
+                    'timestamp' => $envelope->getTimeStamp(),
+                    'priority' => $envelope->getPriority(),
+                    'expiration' => $envelope->getExpiration(),
+                    'app_id' => $envelope->getAppId(),
+                    'message_id' => $envelope->getMessageId(),
+                    'reply_to' => $envelope->getReplyTo(),
+                    'correlation_id' => $envelope->getCorrelationId(),
+                    'headers' => $envelope->getHeaders(),
+                    'user_id' => $envelope->getUserId(),
+                    'cluster_id' => 0,
+                    'channel' => '',
+                    'consumer_tag' => '',
+                ),
+                $envelope->getDeliveryTag()
+            ));
+
+        }, AMQP_NOPARAM, $consumerTag);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     public function get()
     {
         $envelope = $this->queue->get();


krichprollsch:
Hello Olivier,
FYI there was a past discussion about the usage of `consume`  here: #106

makasim:
Just FYI. 

That's something AMQP Interop [already supports](https://github.com/queue-interop/amqp-interop/blob/03cfac42483d07ab45d1896a6a2e1d873a216bba/src/AmqpContext.php#L61-L89). You could you use it as a source of inspiration. 

That could be used even now without any changes to the interfaces (with the [interop borker](https://github.com/swarrot/swarrot/blob/master/src/Swarrot/Broker/MessagePublisher/InteropMessagePublisher.php)). 

```php
<?php
use Enqueue\AmqpExt\AmqpConnectionFactory;
use Swarrot\Broker\MessageProvider\InteropMessageProvider;

$context =  (new AmqpConnectionFactory([
    'dsn' => 'amqp://guest:guest@localhost:5672/%2f'),
    'receive_method' => 'basic_consume',
])->createContext();

$provider = new InteropMessageProvider($context, 'a_queue');

$message = $provider->get();
```

odolbeau:
Even if I'm definitively not a big fan a the `consume` method (because as you said, it's blocking), it looks like there is people out there who wants to use it. :)

Regarding the implementation, it's up to you!
If the `InteropMessageProvider` example provided by @makasim is not enough for you, I'm not against supported it directly in swarrot.

For those who read this thread keep in mind the `consume` method is **blocking**, which means that a lot of processors won't work as expected:
* the `SignalHandlerProcessor`: if you request your processor to stop when it's polling, you'll have to wait until the next message is processed. When you want to stop a consumer which run on an a heavy loaded queue, that's not a big deal. When you want to stop a consumer which run on a n"not so busy" queue, you may have to wait a bit. When you want to stop a consumer which run on an empty queue: you're fucked up!
* the same logic apply for the `MaxExecutionTimeProcessor`. If you consume an empty queue, your processor will never stop.
* the `InsomniacProcessor` won't work at all.

ping @NoiseByNorthwest who may be interested by this thread. :)

NoiseByNorthwest:
Thx @odolbeau !
I still think that all the mentioned issues can be addressed by specifying a timeout. I use this pattern with my beanstalkd workers, 5s timeout for reserve(), which means they can be gracefully terminated at most 5 seconds after signaling them.

makasim:
As @NoiseByNorthwest said timeout solved most of the problems. I'd like to add, bunny and php-amqplib stop consuming on a signal. amqp-ext does not do it and you have to wait till the time is out.

metfan:
I think it's great idea to support consume method. It could be a configurable option. 
I used it with RabbitMQBundle and it allowed me to add some logic on top of the command that I can't achieve with Swarrot.

I don't know how many consumers I have to start because my load is not linear. So I don't want to have 10 consumers when I have 10 messages/min but when I recieved 10k instantly I want a lot of consumers.

With this idea, on previous project I changed the command, at the begining I polled RabbitMQ REST API to know how many messages are in queue and how many consumers already do their jobs. Based on the info I stopped the consumer or let him consumed message.

It could be great if I can do the same with swarrot.

src/Swarrot/Consumer.php
```diff
@@ -63,6 +63,7 @@ public function consume(array $options = array())
         $this->optionsResolver->setDefaults(array(
             'poll_interval' => 50000,
             'queue' => $this->messageProvider->getQueueName(),
+            'method' => 'consume'
```
odolbeau: `'method' => 'get'` by default

src/Swarrot/Consumer.php
```diff
@@ -75,25 +76,54 @@ public function consume(array $options = array())
             $this->processor->initialize($options);
         }
 
+        if ($options['method'] === 'consume') {
+            $this->methodConsume($options);
+        } else {
+            $this->methodGet($options);
+        }
+
+
```
odolbeau: extra blank line

src/Swarrot/Consumer.php
```diff
@@ -75,25 +76,54 @@ public function consume(array $options = array())
...
                     break 2;
                 }
             }
-
             if ($this->processor instanceof SleepyInterface) {
                 if (false === $this->processor->sleep($options)) {
                     break;
                 }
             }
-
```
stof: you should keep these empty lines

src/Swarrot/Consumer.php
```diff
@@ -75,25 +76,54 @@ public function consume(array $options = array())
...
+                }
+
+                if ($this->processor instanceof SleepyInterface) {
+                    if (false === $this->processor->sleep($options)) {
+                        throw new \Exception('Stop processing');
+                    }
+                }
+            });
+
+        } catch (\Exception $e) {}
```
stof: swallowing any exception silently looks very bad. If the processor throws, this will get lost and debugging will not be possible (you probably won't even get an error report, unless your monitoring considers that a consumer exiting with a success exit code is also an issue, but you will still not know what happened).

src/Swarrot/Broker/MessageProvider/MessageProviderInterface.php
```diff
@@ -13,6 +13,14 @@
      */
     public function get();
 
+    /**
+     * consume
```
stof: this phpdoc is totally useless. It should either describe what the method does or be removed (and yes, I know that `get` is already using such useless phpdoc)

src/Swarrot/Broker/MessageProvider/MessageProviderInterface.php
```diff
@@ -13,6 +13,14 @@
...
+     *
+     * @param string $consumerTag Identify the consumer for the rabbitmq-server
```
stof: Talking about `rabbitmq-server` in the description of this interface looks wrong to me. This interface is not specific to RabbitMQ.

stof: The Consumer class currently generates a random value here. If that's the intended usage, PeclPackageMessageProvider should do it internally, removing that argument from the interface IMO.

src/Swarrot/Broker/MessageProvider/MessageProviderInterface.php
```diff
@@ -13,6 +13,14 @@
...
+     * @param callable $callback A callback receiving a Swarrot\Broker\Message instance on consumption
+     */
+    public function consume($consumerTag, callable $callback);
```
stof: adding a new method in an existing interface will force making a new major version of the library, as it is a BC break for any implementor of the interface.

Thus, making it required in the interface would make the interface much harder to implement for any other backend. I suggest making part of a separate optional interface.
And then, handling the case of requesting the `consume` method in Consumer for a provider not implementing that interface could be done in multiple ways:
- throwing an exception saying it is unsupported
- silently switching to `get`
- wrapping it into an adapter which would implement a loop in PHP to emulate `consume` based on `get`. But I think this would simply duplicate what `get` does for no actual benefit, and so is not needed.

Among the existing implementation in the library, I'm quite sure some of them are incompatible with this method.

olaurendeau:
@odolbeau given the review effort of @stof I have to move forward with this !

olaurendeau:
I'm gonna give a try to @makasim option, if it does the job, I'll just propose a README update explaining how to achieve `consume` through queue-interop

makasim:
Some time ago [SubscriptionConsumer](https://github.com/queue-interop/queue-interop/blob/master/src/SubscriptionConsumer.php) interface was introduced. It is a generalized concept of AMQP Interop consumer. 

It is preferred to use subscription consumption over basic get method because: 

* It does not eat resource while waiting for a message
* RabbitMQ QoS feature
* Consume from multiple queues in one process.

makasim:
The comment https://github.com/swarrot/swarrot/pull/169#issuecomment-356886832 is not relevant anymore (as of 0.9 version).  Internally it uses basic.get AMQP call now.

Now, It should be:
```php
<?php
use Enqueue\AmqpExt\AmqpConnectionFactory;

$dsn = 'amqp://guest:guest@localhost:5672/%2f';
$context =  (new AmqpConnectionFactory($dsn)->createContext();

$queue = $context->createQueue('foo');
$consumer = $context->createConsumer($queue);

$subscriptionConsumer = $context->createSubscriptionConsumer();
$subscriptionConsumer->subscripbe($consumer, function(AmqpMessage $message, AmqpConsumer $consumer) {
  // process message from queue foo

  $consumer->acknowledge($message);

  return true;
});

$timeout = 10000; // 10 sec
$subscriptionConsumer->consume($timeout);
```

piotrbrzezina:
@olaurendeau  Any update regarding this PR?

olaurendeau:
Hey Piotr,
I give it a new try a while ago but didn’t succeed to make it happen, two young children are consuming my free time :)
Closing this one.
Feel free to make it move forward !

PR #170, opened 2018-01-12
[ACKProcessor] Add queue name in log context

noniagriconomie:
Hi,

IMO it could be great to have such information in the log context
What do you think?
Also what do you think about rewriting all log message so they are invaraible, and moving variable(s) to the context?

Thank you

### DIFF
--- a/src/Swarrot/Processor/Ack/AckProcessor.php
+++ b/src/Swarrot/Processor/Ack/AckProcessor.php
@@ -52,6 +52,7 @@ class AckProcessor implements ConfigurableInterface
                 '[Ack] Message #'.$message->getId().' has been correctly ack\'ed',
                 [
                     'swarrot_processor' => 'ack',
+                    'queue_name' => $this->messageProvider->getQueueName(),
                 ]
             );
 
@@ -94,6 +95,7 @@ class AckProcessor implements ConfigurableInterface
             ),
             [
                 'swarrot_processor' => 'ack',
+                'queue_name' => $this->messageProvider->getQueueName(),
                 'exception' => $exception,
             ]
         );


odolbeau:
Thanks! 👍

odolbeau:
Thanks for your contribution.

Another solution (used at BlaBlaCar) is to create a dedicated logger for swarrot and a custom processor to add context to this logger (queue & connection names, etc.).
Even if it's sounds more complicated, you'll have those information in all your logs, not only in this one.

@olaurendeau how do you deal with this use case?

noniagriconomie:
Thank you for your reply
I already log all swarrot’s logs in a specific channel and i think the ack processor is one of the most used one
So to add some stat on successful/failing queues i just want to simply have this log context in this processor log message

noniagriconomie:
Any news for this PR? @odolbeau @olaurendeau

Thank you

olaurendeau:
Hello !
In my usecases at LaFourchette so far we never needed to log technical informations about queue, channel, connection, ... What we do log is the routing_key, infos about the message content, consumption time, ...
@odolbeau If I understand well, with the blablacar processor that add context to all the logs, queue name would be added in the AckProcessor logs ?
Could it be a standard Processor ?

noniagriconomie:
@odolbeau thank you for your answer

> Could it be a standard Processor ?

What about doing a DefaultLoggerProcessor?
with some options to enable/disable some data to log

odolbeau:
Sorry for the late answer, I don't have much time to dedicate to this project right now (and with a limited connection).

Anyway, I don't see any elegant solution to implement a system like this by default, mainly because information to log depends on your business (and can be directly in the message itself).

The easiest solution is to merge this PR I think.

@noniagriconomie could you please rebase your branch and check tests?

noniagriconomie:
@odolbeau I will try to review this Pr this week
thank you

noniagriconomie:
hi @odolbeau do you think it is mergeable after helping fixing test?
thank you

odolbeau:
@noniagriconomie Sure it is!
Sorry for the delay, I have a limited internet access for a few weeks.

noniagriconomie:
Closed, we deal differently
Thank you

PR #171, opened 2018-02-15
[XDeathProcessor] Add XDeath max count and max lifetime processor

instabledesign:
|**Q**|**A**|
| ------------- |:-------------:|
| Branch? | master|
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| Tests pass? | yes |

Hi

When we use rabbitMQ retry (DLX, DLK) with nack message we need to correctly handle poison message. 
So there is 2 new middlewares.
 - XDeathMaxCountProcessor
 - XDeathMaxLifetimeProcessor

hope you enjoy :)

### DIFF
--- a/src/Swarrot/Processor/XDeath/README.md
+++ b/src/Swarrot/Processor/XDeath/README.md
@@ -1,0 +1,103 @@
+# XDeathMaxCountProcessor
+
+XDeathMaxCountProcessor is a [swarrot](https://github.com/swarrot/swarrot) processor.
+Its goal is to execute a callback when x-death max count reached in order to kill poison message.
+
+This is a very different approach than RetryProcessor because there is no software clone message republish.
+This processor use the native rabbitmq x-death functionnality.
+
+## Configuration
+
+|Key                                  |Default|Description                                                                  |
+|:-----------------------------------:|:-----:|-----------------------------------------------------------------------------|
+|x_death_max_count                    |300    |The number of attempts before callback.
+|x_death_max_count_log_levels_map     |[]     |Map of classes to a log level when retry. (Warning by default)               |
+|x_death_max_count_fail_log_levels_map|[]     |Map of classes to a log level when all retries failed. (Warning by default)  |
+
+## How it works
+
+When your processor throw an exception (= failed to process a message), the
+XDeathMaxCount Processor will catch it and execute the callback if the message header x-death count
+exceed the `x_death_max_count`.
+
+/!\ Caution ! If your callback returns null, the exception will be rethrown.
+
+## Real example
+
+Let's say you want to consume a queue named `mail`. You would like to have 3
+retries if the mail is not sent correctly. Each retry was separate by 30 seconds.
+
+First let's create the delaying exchange+queue 
+
+* Create an exchange `waiting_30` (type: `topic`)
+* Create an queue `waiting_30` (x-dead-letter-exchange: ` `, x-message-ttl: `30000`)
+* Bind the exchange `waiting_30` to the queue `waiting_30` with routing_key `#`
+
+And then create the simple mail workflow.
+
+* Create an exchange `mail` (type: `direct`)
+* Create an queue `queue_mail` with (x-dead-letter-exchange: `waiting_30`, x-dead-letter-routing-key: `queue_mail`)
+* Bind the exchange `mail` to the queue `queue_mail` with routing_key `queue_mail`
+
+You must use the AckProcessor (or nack the message by yourself).
+
+Once this configuration is done, when an exception is thrown, the
+AckProcessor will nack the message and rabbitmq will append or increase message header
+x-death count. This x-death header is the message trace when the message was `reject` or `expired` by a queue.
+So the message was reject to exchange `waiting_30` and route to queue `waiting_30`.
+After 30 seconds the queue `expire` the message and move it to AMQP default exchange
+and route directly to queue `queue_mail`.
+If your consumer throw exception 3 times then the XDeathMaxCount will execute the configured callback 
+and not rethrow the exception in order to let the Ack processor ack the message to stop retrying it.
+
+
+# XDeathMaxLifetimeProcessor
+
+XDeathMaxLifetimeProcessor is a [swarrot](https://github.com/swarrot/swarrot) processor.
+Its goal is to execute a callback when x-death time exceed max lifetime in order to kill poison message.
+
+This is a very different approach than RetryProcessor because there is no software clone message republish.
+This processor use the native rabbitmq x-death functionnality.
+
+## Configuration
+
+|Key                                     |Default|Description                                                                  |
+|:--------------------------------------:|:-----:|-----------------------------------------------------------------------------|
+|x_death_max_lifetime                    |3600   |The number seconds before callback.
+|x_death_max_lifetime_log_levels_map     |[]     |Map of classes to a log level when retry. (Warning by default)               |
+|x_death_max_lifetime_fail_log_levels_map|[]     |Map of classes to a log level when all retries failed. (Warning by default)  |
+
+## How it works
+
+When your processor throw an exception (= failed to process a message), the
+XDeathMaxLifetime Processor will catch it and execute the callback if the message header x-death time
+exceed the `x_death_max_lifetime`.
+
+/!\ Caution ! If your callback returns null, the exception will be rethrown.
+
+## Real example
+
+Let's say you want to consume a queue named `mail`. You would like to
+retry during 1 hour if the mail is not sent correctly. Each retry was separate by 30 seconds.
+
+First let's create the delaying exchange+queue 
+
+* Create an exchange `waiting_30` (type: `topic`)
+* Create an queue `waiting_30` (x-dead-letter-exchange: ` `, x-message-ttl: `30000`)
+
+And then create the simple mail workflow.
+
+* Create an exchange `mail` (type: `direct`)
+* Create an queue `queue_mail` with (x-dead-letter-exchange: `waiting_30`, x-dead-letter-routing-key: `queue_mail`)
+* Bind the exchange `mail` to the queue `queue_mail` with routing_key `queue_mail`
+
+You must use the AckProcessor (or nack the message by yourself).
+
+Once this configuration is done, when an exception is thrown, the
+AckProcessor will nack the message and rabbitmq will append x-death time header (with timestamp). 
+This x-death header is the message trace when the message was `reject` or `expired` by a queue.
+So the message was reject to exchange `waiting_30` and route to queue `waiting_30`.
+After 30 seconds the queue `expire` the message and move it to AMQP default exchange
+and route directly to queue `queue_mail`.
+After 1 hour if your consumer throw exception then the XDeathMaxLifetime will execute the configured callback 
+and not rethrow the exception in order to let the Ack processor ack the message to stop retrying it.
--- a/src/Swarrot/Processor/XDeath/XDeathMaxCountProcessor.php
+++ b/src/Swarrot/Processor/XDeath/XDeathMaxCountProcessor.php
@@ -1,0 +1,176 @@
+<?php
+
+namespace Swarrot\Processor\XDeath;
+
+use PhpAmqpLib\Wire\AMQPArray;
+use Psr\Log\LoggerInterface;
+use Psr\Log\LogLevel;
+use Psr\Log\NullLogger;
+use Swarrot\Broker\Message;
+use Swarrot\Processor\ProcessorInterface;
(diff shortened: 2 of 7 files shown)


src/Swarrot/Processor/XDeath/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Real example
+
+Let's say you want to consume a queue named `mail`. You would like to have 3
+retries if the mail is not sent correctly. Each retry was separate by 30 seconds.
+
+First let's create the delaying exchange+queue 
+
+* Create an exchange `waiting_30` (type: `topic`)
+* Create an queue `waiting_30` (x-dead-letter-exchange: ``, x-message-ttl: `30000`)
```
stof: The 2 backticks look weird here. They are not creating an inline docblock in the markdown rendering. What are you trying to represent here ?

src/Swarrot/Processor/XDeath/README.md
```diff
@@ -0,0 +1,103 @@
...
+* Create an exchange `mail` (type: `direct`)
+* Create an queue `queue_mail` with (x-dead-letter-exchange: `waiting_30`, x-dead-letter-routing-key: `queue_mail`)
+* Bind the exchange `mail` to the queue `queue_mail` with routing_key `queue_mail`
+
+You must use the AckProcessor (or nack the message by yourself).
+
+Once this configuration is done, when an exception is thrown, the
+AckProcessor will nack the message and rabbitmq will append or increase message header
+x-death count. This x-death header is the message trace when the message was `reject` or `expired` by a queue.
+So the message was rely to exchange `waiting_30` and route to queue `waiting_30`.
```
stof: `was rely` ? This sentence looks weird to me; Are you sure about it ?

src/Swarrot/Processor/XDeath/XDeathMaxCountProcessor.php
```diff
@@ -0,0 +1,180 @@
...
+     *
+     * @return mixed
+     */
+    private function handleException($exception, Message $message, array $options)
+    {
+        $headers = $message->getProperties();
+        if (isset($headers['headers']['x-death'], $options['x_death_max_count'])) {
+            $xDeathHeaders = $headers['headers']['x-death'];
+            // PhpAmqpLib compatibility
+            if (class_exists('PhpAmqpLib\Wire\AMQPArray') && $headers['headers']['x-death'] instanceof AMQPArray) {
```
stof: `class_exists` is useless here. `instanceof` is perfectly fine if the class name does not exist (except that you won't be able to make it return `true` in such case)

src/Swarrot/Processor/XDeath/XDeathMaxCountProcessor.php
```diff
@@ -0,0 +1,180 @@
...
+                        $exception,
+                        sprintf(
+                            '[XDeathMaxCount] Max count reached. %d/%d attempts. Execute the configured callback.',
+                            $queueXDeathHeader['count'],
+                            $options['x_death_max_count']
+                        ),
+                        $options['x_death_max_count_fail_log_levels_map']
+                    );
+
+                    if (null !== $return = call_user_func($this->callback, $exception, $message, $options)) {
```
stof: please use a fully-qualified call here (i.e. `\call_user_func`), which will allow recent PHP versions to optimize this call by avoiding to create a method call for `call_user_func` itself (replacing it by a function call at compile time)

src/Swarrot/Processor/XDeath/README.md
```diff
@@ -0,0 +1,103 @@
...
+|x_death_max_lifetime_log_levels_map     |[]     |Map of classes to a log level when retry. (Warning by default)               |
+|x_death_max_lifetime_fail_log_levels_map|[]     |Map of classes to a log level when all retries failed. (Warning by default)  |
+
+## How it works
+
+When your processor throw an exception (= failed to process a message), the
+XDeathMaxLifetime Processor will catch it and execute the callback if the message header x-death time
+exceed the `x_death_max_lifetime`.
+
+/!\ Caution ! While your max lifetime callback return null the exception gonna to be rethrow
```
stof: this sentences looks weird to me. What about ``If your callback returns `null`, the exception will be rethrown.` ?

And in any case, the period is missing at the end of the sentence.

src/Swarrot/Processor/XDeath/XDeathMaxCountProcessor.php
```diff
@@ -0,0 +1,180 @@
...
+use Swarrot\Processor\ProcessorInterface;
+use Swarrot\Processor\ConfigurableInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+class XDeathMaxCountProcessor implements ConfigurableInterface
+{
+    /**
+     * @var ProcessorInterface
+     */
+    protected $processor;
```
stof: I would make it private too

src/Swarrot/Processor/XDeath/XDeathMaxCountProcessor.php
```diff
@@ -0,0 +1,180 @@
...
+    /**
+     * @param ProcessorInterface $processor
+     * @param MessageProviderInterface $messageProvider
+     * @param \Closure $callback
+     * @param LoggerInterface|null $logger
+     */
+    public function __construct(
+        ProcessorInterface $processor,
+        MessageProviderInterface $messageProvider,
+        \Closure $callback,
```
stof: should be `callable` rather than `\Closure`. I don't see any benefit restricting the kind of callable you can use (btw, when using the Symfony DI container, forcing to use closures would make it harder to configure, because you cannot directly put a closure in the container)

src/Swarrot/Processor/XDeath/XDeathMaxCountProcessor.php
```diff
@@ -0,0 +1,180 @@
...
+     * @param $exception
+     * @param Message $message
+     * @param array $options
+     *
+     * @return mixed
+     */
+    private function handleException($exception, Message $message, array $options)
+    {
+        $headers = $message->getProperties();
+        if (isset($headers['headers']['x-death'], $options['x_death_max_count'])) {
```
stof: no need to check whether `$options['x_death_max_count']` is set. It will always be set when using the processor in a supported way, as a default value is defined for the option

src/Swarrot/Processor/XDeath/XDeathMaxCountProcessor.php
stof: you could use the `$xDeathHeaders` defined just above here.

src/Swarrot/Processor/XDeath/XDeathMaxLifetimeProcessor.php
```diff
@@ -0,0 +1,186 @@
...
+use Swarrot\Processor\ProcessorInterface;
+use Swarrot\Processor\ConfigurableInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+class XDeathMaxLifetimeProcessor implements ConfigurableInterface
+{
+    /**
+     * @var ProcessorInterface
+     */
+    protected $processor;
```
stof: should be private

src/Swarrot/Processor/XDeath/XDeathMaxLifetimeProcessor.php
```diff
@@ -0,0 +1,186 @@
...
+
+    /**
+     * @var LoggerInterface
+     */
+    private $logger;
+
+    /**
+     * @param ProcessorInterface $processor
+     * @param MessageProviderInterface $messageProvider
+     * @param \Closure $callback
```
stof: same here about the callable type

src/Swarrot/Processor/XDeath/XDeathMaxLifetimeProcessor.php
```diff
@@ -0,0 +1,186 @@
...
+    private function handleException($exception, Message $message, array $options)
+    {
+        $headers = $message->getProperties();
+        if (isset($headers['headers']['x-death'], $options['x_death_max_lifetime'])) {
+            $xDeathHeaders = $headers['headers']['x-death'];
+            // PhpAmqpLib compatibility
+            if (class_exists('PhpAmqpLib\Wire\AMQPArray') && $headers['headers']['x-death'] instanceof AMQPArray) {
+                $xDeathHeaders = $headers['headers']['x-death']->getNativeData();
+            }
+            $queueName = $this->messageProvider->getQueueName();
```
stof: why injecting the whole MessageProvider just to get the queue name ? You could inject just the queue name

src/Swarrot/Processor/XDeath/XDeathMaxLifetimeProcessor.php
```diff
@@ -0,0 +1,186 @@
...
+                    $exception,
+                    sprintf(
+                        '[XDeathMaxLifetime] No x-death header found for queue name "%s". Do nothing.',
+                        $queueName
+                    ),
+                    $options['x_death_max_lifetime_fail_log_levels_map']
+                );
+            }
+
+            if (isset($queueXDeathHeader, $queueXDeathHeader['time'])) {
```
stof: use `null !== $queueXDeathHeader` instead of putting it inside the `isset`. It makes the code more robust by forbidding undefined variable.

src/Swarrot/Processor/XDeath/XDeathMaxLifetimeProcessor.php
```diff
@@ -0,0 +1,186 @@
...
+                        $exception,
+                        sprintf(
+                            '[XDeathMaxLifetime] Max lifetime reached. %s/%s seconds exceed. Execute the configured callback.',
+                            abs($remainLifetime),
+                            $options['x_death_max_lifetime']
+                        ),
+                        $options['x_death_max_lifetime_fail_log_levels_map']
+                    );
+
+                    if (null !== $return = call_user_func($this->callback, $exception, $message, $options)) {
```
stof: same here about the `\call_user_func` call

tests/Swarrot/Processor/XDeath/XDeathMaxCountProcessorTest.php
```diff
@@ -0,0 +1,398 @@
+<?php
+
+namespace Swarrot\Tests\Processor\Retry;
```
stof: the namespace is wrong

tests/Swarrot/Processor/XDeath/XDeathMaxLifetimeProcessorTest.php
```diff
@@ -0,0 +1,427 @@
+<?php
+
+namespace Swarrot\Tests\Processor\Retry;
```
stof: wrong namespace

src/Swarrot/Processor/XDeath/README.md
instabledesign: an empty string (to bind the queue to the AMQP Default exchange)

src/Swarrot/Processor/XDeath/XDeathMaxCountProcessor.php
instabledesign: i set the class_exists because the tests was brake without it.
have you another solution?

src/Swarrot/Processor/XDeath/XDeathMaxCountProcessor.php
instabledesign: :+1:

src/Swarrot/Processor/XDeath/README.md
stof: For things to work, you need to put a space between the start and the end delimiter of the code span (the content between delimiter is trimmed, so you will get an empty string in the end, but it is needed to separate them to create a code span)

src/Swarrot/Processor/XDeath/XDeathMaxCountProcessor.php
stof: are you sure about *this* `class_exists` ? In the commit after that, you are adding one in the test data provider, and this one is indeed legitimate (as it wraps a *usage* of the class). But here, it should not be necessary.

stof:
please also fix the coding standard to make the CI happy about them.

instabledesign:
*Thx for your review.*
It's all ok for me. 
I search how to implement his configurator in the bundle part (because of callable injection).

instabledesign:
It's ready to merge. 
@odolbeau what did you think about it.

instabledesign:
or @stof

instabledesign:
reping @odolbeau @stof

odolbeau:
Really good job!
Looks good to me and it looks like you corrected everything @stof pointed out.

Let's merge!

Issue #137, opened 2016-11-14
RetryProcessor message is not compatible with Amqplib >= 2.5


pdoreau:
Hi,
An exception is thrown when the second retry message is given to Amqplib (>=2.5) :

```
[Retry] An exception occurred. Republish message for the 2 times ...
                                                         
  [Symfony\Component\Debug\Exception\ContextErrorException]            
  Warning: mb_strlen() expects parameter 1 to be string, object given                                                                   

Exception trace:
 () at /app/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPWriter.php:350
 Symfony\Component\Debug\ErrorHandler->handleError() at n/a:n/a
 mb_strlen() at /app/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPWriter.php:350
 PhpAmqpLib\Wire\AMQPWriter->write_longstr() at /app/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPWriter.php:477
 PhpAmqpLib\Wire\AMQPWriter->write_value() at /app/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPWriter.php:410
 PhpAmqpLib\Wire\AMQPWriter->write_table() at /app/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPWriter.php:425
 PhpAmqpLib\Wire\AMQPWriter->write_table_object() at /app/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/GenericContent.php:191
 PhpAmqpLib\Wire\GenericContent->serialize_properties() at /app/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Channel/AMQPChannel.php:1099
 PhpAmqpLib\Channel\AMQPChannel->basic_publish() at /app/vendor/swarrot/swarrot/src/Swarrot/Broker/MessagePublisher/PhpAmqpLibMessagePublisher.php:46
 Swarrot\Broker\MessagePublisher\PhpAmqpLibMessagePublisher->publish() at /app/vendor/swarrot/swarrot/src/Swarrot/Processor/Retry/RetryProcessor.php:121
 Swarrot\Processor\Retry\RetryProcessor->handleException() at /app/vendor/swarrot/swarrot/src/Swarrot/Processor/Retry/RetryProcessor.php:33
[...]

PR #174, opened 2018-05-03
[PhpAmqpLibPublisher] Misbehaving header processing with PhpAmqpLib\AMQPArray

claire-lovisa:
Hi!

**Description**

When using the RetryProcessor with PhpAmqpLibPublisher, I had the following error after the second retry :

```
mb_strlen() expects parameter 1 to be string, object given in /app/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPWriter.php on line 365
``` 

Related issue : https://github.com/swarrot/swarrot/issues/137

You can reproduce it with the files here : https://gist.github.com/claire-lovisa/724d5480ed4c2d812acef9faf0014523 _(versions used : php-amqp-lib 2.7, swarrot 3.2, rabbitmq 3.7)_

**Explanation**

It was due to a wrong type attribution when processing the headers given by RabbitMQ (x-death) after the first retry. In fact, PhpAmqpLib can give AMQPArray to Swarrot in the message headers and Swarrot attributed to it the 'S' type because AMQPArray != array. PhpAmqpLib\AMQPWriter was then not able to process the AMQPArray as a string.

Here's the way I fixed it, I hope it might be interesting for you too :)

### DIFF
--- a/src/Swarrot/Broker/MessagePublisher/PhpAmqpLibMessagePublisher.php
+++ b/src/Swarrot/Broker/MessagePublisher/PhpAmqpLibMessagePublisher.php
@@ -58,7 +58,7 @@ class PhpAmqpLibMessagePublisher implements MessagePublisherInterface
                 $properties['application_headers'] = [];
             }
             foreach ($properties['headers'] as $header => $value) {
-                if (is_array($value) or ($value instanceof AMQPArray)) {
+                if (is_array($value) || $value instanceof AMQPArray) {
                     $type = 'A';
                 } elseif (is_int($value)) {
                     $type = 'I';
--- a/src/Swarrot/Broker/MessagePublisher/PhpAmqpLibMessagePublisher.php
+++ b/src/Swarrot/Broker/MessagePublisher/PhpAmqpLibMessagePublisher.php
@@ -4,6 +4,7 @@ namespace Swarrot\Broker\MessagePublisher;
 
 use PhpAmqpLib\Message\AMQPMessage;
 use PhpAmqpLib\Channel\AMQPChannel;
+use PhpAmqpLib\Wire\AMQPArray;
 use Swarrot\Broker\Message;
 
 class PhpAmqpLibMessagePublisher implements MessagePublisherInterface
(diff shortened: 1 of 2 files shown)


src/Swarrot/Broker/MessagePublisher/PhpAmqpLibMessagePublisher.php
```diff
@@ -57,7 +58,7 @@ public function publish(Message $message, $key = null)
                 $properties['application_headers'] = [];
             }
             foreach ($properties['headers'] as $header => $value) {
-                if (is_array($value)) {
+                if (is_array($value) or ($value instanceof AMQPArray)) {
```
odolbeau: Prefer: `if (is_array($value) || $value instanceof AMQPArray) {`

odolbeau:
Hi @claire-lovisa (and sorry for the delay).

Thanks a lot for this PR, nice catch. :)
It's definitively interesting for us too!

Could you please update the test in order to skip it when the `AMQPArray` class is not found?

Thanks again.

claire-lovisa:
Hi @odolbeau !

Great then :smiley: 

The tests are updated. I don't skip the test is the class is not found because you would lose a valuable test (apart from this specific case). Instead, I add a few lines to test AMQPArray when the class is found. Is it OK for you ?

odolbeau:
That's great! :)

Let's merge. :+1:

stof:
I think a better fix would be to unwrap the AMQPArray in the provider, so that headers are always an array, and not a AMQPArray in the Swarrot message.

odolbeau:
Makes sense.
WDYT @claire-lovisa?

claire-lovisa:
Yes that's a good idea! I see how to do it, but @stof do you prefer to do it yourself?

stof:
@claire-lovisa if you have time to do it, please send a PR. I won't have time to do it myself in the next coming days anyway.

instabledesign:
Yeah nice catch im not see this one coming 👍

PR #175, opened 2018-05-25
[PhpAmqpLibPublisher][Fix 2.0] Misbehaving header processing with PhpAmqpLib\AMQPArray

claire-lovisa:
This PR is the continuation of [#174](https://github.com/swarrot/swarrot/pull/174).

As suggested by @stof, instead of correcting the publisher, we can fix the provider so that we always have array in the Swarrot message, not sometimes an array and sometimes an AMQPArray. 

I removed here the previous fix (and corresponding test) and then I fixed the provider and added a test case.

### DIFF
--- a/src/Swarrot/Broker/MessageProvider/PhpAmqpLibMessageProvider.php
+++ b/src/Swarrot/Broker/MessageProvider/PhpAmqpLibMessageProvider.php
@@ -55,8 +55,8 @@ class PhpAmqpLibMessageProvider implements MessageProviderInterface
         $properties['headers'] = [];
         if ($envelope->has('application_headers')) {
             foreach ($envelope->get('application_headers') as $key => $value) {
-                if (is_a($value, AMQPArray::class)) {
-                    $properties['headers'][$key] = $value->getNativeData()[0];
+                if (is_a($value[1], AMQPArray::class)) {
+                    $properties['headers'][$key] = $value[1]->getNativeData();
                 } else {
                     $properties['headers'][$key] = $value[1];
                 }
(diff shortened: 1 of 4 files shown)


tests/Swarrot/Broker/MessageProvider/PhpAmqpLibMessageProviderTest.php
```diff
@@ -25,6 +26,36 @@ public function test_get_with_messages_in_queue_return_message()
         $this->assertInstanceOf('Swarrot\Broker\Message', $message);
     }
 
+    public function test_get_with_amqp_array_header_return_array_header()
+    {
+        if (class_exists(AMQPArray::class)) {
```
odolbeau: It's better to mark test as skipped when the class doesn't exists.

it's easier to read than a huge if and it won't display a notice for an empty test if the class is not found.

src/Swarrot/Broker/MessageProvider/PhpAmqpLibMessageProvider.php
```diff
@@ -54,7 +55,11 @@ public function get()
         $properties['headers'] = [];
         if ($envelope->has('application_headers')) {
             foreach ($envelope->get('application_headers') as $key => $value) {
-                $properties['headers'][$key] = $value[1];
+                if (is_a($value[1], AMQPArray::class)) {
```
odolbeau: Can you use `instanceof` instead of `is_a` please?

odolbeau:
Thanks! :+1:

PR #180, opened 2018-09-19
Add possibility to define ack and nack handlers with publisher confirms

julienthomas:
I needed some way to customize the handlers for publisherConfirm.

I put these modifications directly on the publish method to be able to use this with swarrot bundle

### DIFF
--- a/src/Swarrot/Broker/MessagePublisher/InteropMessagePublisher.php
+++ b/src/Swarrot/Broker/MessagePublisher/InteropMessagePublisher.php
@@ -38,7 +38,7 @@ final class InteropMessagePublisher implements MessagePublisherInterface
     }
 
     /** {@inheritdoc} */
-    public function publish(Message $message, $key = null)
+    public function publish(Message $message, $key = null, callable $ackHandler = null, callable $nackHandler = null)
     {
         $headers = $message->getProperties();
         $properties = [];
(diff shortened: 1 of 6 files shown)


odolbeau:
Hi @julienthomas & thanks for your contribution!

The Ack/Nack logic is not something common to all brokers out there and that's not something everyone use.
The goal of swarrot is to provide an "on-demand" library where you can activate (or not) any functionality provided by the library.

The Ack/Nack logic is provided by the [AckProcessor](https://github.com/swarrot/swarrot/blob/master/src/Swarrot/Processor/Ack/AckProcessor.php). You can use it easily by creating a stack [like described in the README](https://github.com/swarrot/swarrot#using-a-stack)

```php
    ->push('Swarrot\Processor\Ack\AckProcessor', $messageProvider)
;

```

Tell me if it respond to your use case.

julienthomas:
Hi.

I didn't read that the processors could be used with a publisher.

In my case I want rabitMQ to send a basic.nack to my consumer using queue length limit (rabbit 3.7)

julienthomas:
Hi.

Not if I want to use custom handlers from SarrotBundle because we don't have any access to these constructors.

But we can remove the interface modification, php allow class that implements that interface to have a different method signature if the added parameters to the method have a default value.

odolbeau:
You don't have any access **yet**.
Choosing an implementation which isn't the best one in this library cause you won't be able to use it out of the box with the bundle is definitively not what we want.

I think you should definitively add those handlers in the publisher constructor (you don't have to do it for all of them, you can just update the publisher you're interested in) and then work on a bundle improvement.

julienthomas:
Ok, I will try to modify this.

Thank you for your answers.

PR #185, opened 2019-03-12
Upgrade ext-amqp on CI

stof:
The `php: nightly` jobs were failing to compile ext-amqp. This is an attempt at fixing it (it should at least allow having a working PHP 7.3 job when we add it, if not nightly)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ addons:
 env:
     global:
         - LIBRABBITMQ_VERSION=v0.8.0
-        - PHPAMQP_VERSION=v1.9.3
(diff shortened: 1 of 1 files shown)


stof:
OK, `php: nightly` still fails to compile, because nightly is 8.0.0-dev. But this PR allowed me to add testing on PHP 7.3

odolbeau:
We can even remove php nightly support for now, it doesn't make sense to have it WDYT?

stof:
yeah, makes sense

stof:
@odolbeau removed

odolbeau:
:+1:

PR #186, opened 2019-04-29
Fixed SignalHandlerProcessor

lyrixx:
Before the patch, the signal handler where unregister in the shouldStop method. so basically ... it does not work

### DIFF
--- a/src/Swarrot/Processor/SignalHandler/SignalHandlerProcessor.php
+++ b/src/Swarrot/Processor/SignalHandler/SignalHandlerProcessor.php
@@ -109,11 +109,6 @@ class SignalHandlerProcessor implements ConfigurableInterface, SleepyInterface, 
     {
         pcntl_signal_dispatch();
 
-        $signals = isset($options['signal_handler_signals']) ? $options['signal_handler_signals'] : array();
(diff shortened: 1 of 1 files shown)


odolbeau:
Nice catch, thanks!
I think you can simply remove those lines, removing the custom signal handler doesn't makes sense IMO.

lyrixx:
@odolbeau Done :ok_hand:

src/Swarrot/Processor/SignalHandler/SignalHandlerProcessor.php
```diff
@@ -67,7 +63,7 @@ public function process(Message $message, array $options)
     {
         $return = $this->processor->process($message, $options);
 
-        if (extension_loaded('pcntl') && $this->shouldStop($options)) {
```
lyrixx: I don't understand this code. IMHO this is wrong. We should always return `$return`, isn't ?

odolbeau: Right!

odolbeau:
Thanks for those nice improvements!

PR #188, opened 2019-05-16
Use static log messages

### LINKED ISSUE #187: Avoid dynamic messages when logging
Right now, when I look at my Kibana dashboard, I get no grouping for some messages:

kibana

This can easily be fixed by using static messages and a dynamic `$context`:

```php
    $this->logger->info(
        "[Ack] A message has been correctly ack'ed",
        [
           'message_id' => $message->getId(),
           'swarrot_processor' => 'ack',
        ]
    );
```

instead of https://github.com/swarrot/swarrot/blob/4b945ad0225a1be8e275ee83849482f3a39f33b3/src/Swarrot/Processor/Ack/AckProcessor.php#L51-L56

If you agree, I can create a PR. There are many more instances of that issue, that's why I'm asking before fixing them all.

greg0ire:
This plays better with logging systems: those will often group logs by
message, showing how many time similar events happened.

Fixes #187

Not using [placeholders](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-3-logger-interface.md#12-message) for now because I don't know how widely used they are, tell me if you want me to.

I fixed 2 out 18 occurrences for now, but I like early feedback.

EDIT: 18/18

### DIFF
--- a/src/Swarrot/Processor/Ack/AckProcessor.php
+++ b/src/Swarrot/Processor/Ack/AckProcessor.php
@@ -49,8 +49,9 @@ class AckProcessor implements ConfigurableInterface
             $this->messageProvider->ack($message);
 
             $this->logger->info(
-                '[Ack] Message #'.$message->getId().' has been correctly ack\'ed',
+                "[Ack] Message has been correctly ack'ed",
                 [
+                    'message_id' => $message->getId(),
                     'swarrot_processor' => 'ack',
                 ]
             );
@@ -87,12 +88,10 @@ class AckProcessor implements ConfigurableInterface
         $this->messageProvider->nack($message, $requeue);
 
         $this->logger->error(
-            sprintf(
-                '[Ack] An exception occurred. Message #%d has been %s.',
-                $message->getId(),
-                $requeue ? 'requeued' : 'nack\'ed'
-            ),
+            '[Ack] An exception occurred.',
             [
+                'message_id' => $message->getId(),
+                'message_status' => $requeue ? 'requeued' : "nack'ed",
                 'swarrot_processor' => 'ack',
                 'exception' => $exception,
             ]
--- a/src/Swarrot/Processor/MemoryLimit/MemoryLimitProcessor.php
+++ b/src/Swarrot/Processor/MemoryLimit/MemoryLimitProcessor.php
@@ -40,8 +40,9 @@ class MemoryLimitProcessor implements ConfigurableInterface
 
         if (null !== $options['memory_limit'] && memory_get_usage() >= $options['memory_limit'] * 1024 * 1024) {
             $this->logger->info(
-                sprintf('[MemoryLimit] Memory limit has been reached (%d MB)', $options['memory_limit']),
+                '[MemoryLimit] Memory limit has been reached (%d MB)',
                 [
+                    'memory_limit' => $options['memory_limit'],
                     'swarrot_processor' => 'memory_limit',
                 ]
             );
--- a/src/Swarrot/Consumer.php
+++ b/src/Swarrot/Consumer.php
@@ -55,10 +55,9 @@ class Consumer
      */
     public function consume(array $options = array())
     {
-        $this->logger->debug(sprintf(
-            'Start consuming queue %s.',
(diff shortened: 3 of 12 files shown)


odolbeau:
I thought you were talking about the `AckProcessor` only! :o 
Given my usage it looks fine for me.

WDYT @stof @olaurendeau @lyrixx?

BTW, placeholders are very useful to have an information in both the message & the context. In your case, it doesn't make sense as you won't have the same message for all logs.

lyrixx:
I like this move for level >= error. For lower it does not worth it imho

greg0ire:
@lyrixx please have a look at the linked issue, the level for which I am the most interested in this is `info`. I want to know how many messages were correctly ack'ed, and I want to do so easily (yes, we log that level in my company). Please reconsider :pray:

odolbeau:
I'm 100% in favor of removing the messageId from the message (for all level) as it's not useful at all.
I'm not sure for other information (for example the queue name).

TBH I don't use any modern log stack for now so I'm not really aware of what are best practices in this domain.

greg0ire:
> TBH I don't use any modern log stack for now so I'm not really aware of what are best practices in this domain.

From the author of "laisse pas trainer ton log" :trollface: 
How can you make such a great presentation and then not use a modern log stack :sweat_smile: ?

lyrixx:
@greg0ire really strange way to monitor rabbitmq !

greg0ire:
@lyrixx we're monitoring the whole app, the goal is not really to monitor RMQ, but to make RMQ-related messages a bit less noisy, so that other messages can be seen too (while still being able to see RMQ-related messages).

odolbeau:
@greg0ire As I'm not working in a company which can afford a complete log stack, I deal with papertrail only (and it's far, far, far from perfect. ^^)

Anyway, is there ano other message to update, or is this PR ready to be merged?

greg0ire:
It's ready, some of the 18 occurences were already fine :)

odolbeau:
Perfect, thanks a lot. :)

src/Swarrot/Processor/Ack/AckProcessor.php
```diff
@@ -87,12 +88,10 @@ private function handleException($exception, Message $message, array $options)
         $this->messageProvider->nack($message, $requeue);
 
         $this->logger->error(
-            sprintf(
-                '[Ack] An exception occurred. Message #%d has been %s.',
-                $message->getId(),
-                $requeue ? 'requeued' : 'nack\'ed'
-            ),
+            '[Ack] An exception occurred.',
```
stof: Maybe we should keep the part about `Message has been requeued|nack'ed.`. There will only be 2 alternatives in the grouping.

src/Swarrot/Processor/MemoryLimit/MemoryLimitProcessor.php
```diff
@@ -40,8 +40,9 @@ public function process(Message $message, array $options)
 
         if (null !== $options['memory_limit'] && memory_get_usage() >= $options['memory_limit'] * 1024 * 1024) {
             $this->logger->info(
-                sprintf('[MemoryLimit] Memory limit has been reached (%d MB)', $options['memory_limit']),
+                '[MemoryLimit] Memory limit has been reached (%d MB)',
```
stof: `\d MB` should be removed, as this is not a sprintf pattern anymore.

src/Swarrot/Processor/Ack/AckProcessor.php
greg0ire: Yeah maybe :thinking:

greg0ire: I will create a separate PR for both issues

src/Swarrot/Processor/MemoryLimit/MemoryLimitProcessor.php
greg0ire: See #191

src/Swarrot/Processor/Ack/AckProcessor.php
greg0ire: See #192

PR #198, opened 2019-12-06
Use github action for php-cs-fixer

odolbeau:
& remove unused phpstan file


Makefile
```diff
@@ -0,0 +1,7 @@
+.PHONY: static
+
+DIR := ${CURDIR}
+QA_IMAGE := jakzal/phpqa:php7.3-alpine
+
+static:
```
stof: I would rename this, as it is checking coding standards, not doing static analysis of the code.

PR #202, opened 2020-01-22
Deprecate Stomp message providers

odolbeau:
Hi @instabledesign 

In order to limit dependencies used by swarrot & improve maintainability, I consider deprecate some processors & providers (see #197 )

Do you still use swarrot & stomp providers?

Thanks!

PR #204, opened 2020-01-22
Fix some type related errors

NicolasCARPi:
* the $key argument of publish() in MessagePublisherInterface should be
  a string
* getExchangeName() in SimpleStompMessagePublisher should return a
  string as per interface contract
* newrelic_notice_error takes a string or an exception as parameter, not
  null and then an exception. See
https://docs.newrelic.com/docs/agents/php-agent/php-agent-api/newrelic_notice_error

### DIFF
--- a/src/Swarrot/Broker/MessagePublisher/MessagePublisherInterface.php
+++ b/src/Swarrot/Broker/MessagePublisher/MessagePublisherInterface.php
@@ -12,7 +12,7 @@ interface MessagePublisherInterface
      * @param Message $message The message to publish
      * @param string  $key     A routing key to use
      */
-    public function publish(Message $message, $key = null);
+    public function publish(Message $message, $key = '');
 
     /**
      * getExchangeName.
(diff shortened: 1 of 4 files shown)


odolbeau:
Hi @NicolasCARPi & tahnks for your contribution.

The `SimpleStompMessageProvider` have been deprecated today (see #202) in order to reduce dependencies of this repository.
If you're actively using it, I suggest you to move this provider in its own repo (managed by you) or we could consider creating a new repository under the swarrot organization if you're willing to help to maintain it.

Regarding the correction of the `newrelic_notice_error` call, don't hesitate to create a dedicated PR in order to allow us to merge & release it asap. :)

For others changes, I'm currently working on a v4.0.0 version which will include strict type hinting everywhere in order to avoid those confusions between null & empty strings. :)

NicolasCARPi:
Salut Olivier,

Oh ok, that's my luck, something got deprecated today! :D 

I have created a dedicated PR for the newrelic issue.

Note that I'm not using the software, so I won't be interested in any sort of fork/dedicated repo.

And yes, strict type hinting is the way to go! ;)

@+
~Nico

PR #205, opened 2020-01-22
Fix newrelic_notice_error() call

NicolasCARPi:
First argument should be a string or a Throwable|Exception. See:
https://docs.newrelic.com/docs/agents/php-agent/php-agent-api/newrelic_notice_error


odolbeau:
Perfect, thanks! :)

PR #208, opened 2020-01-23
Deprecate NewRelicProcessor

odolbeau:
To be removed in 4.0


stof:
Same comment than in #209 : it would be great to explain the deprecation (this is especially helpful for people using this processor)

PR #209, opened 2020-01-23
More deprecation

odolbeau:
Deprecate stomp message publishers & interop message publishers / providers


stof:
What is the reason for this deprecation ? There is no info in this PR and I cannot find any issue giving them either.
I think we need to communicate the reason.

PR #212, opened 2020-01-23
Release v4.0.0

odolbeau:
Prepare release v4.0.0

To be merged after #211


stof:
if the driver classes are removed in 4.0, they should also be marked as deprecated in 3.7

src/Swarrot/Processor/ConfigurableInterface.php
```diff
@@ -6,8 +6,5 @@
 
 interface ConfigurableInterface extends ProcessorInterface
 {
-    /**
-     * setDefaultOptions.
```
stof: We should probably add `@return void` in the phpdoc of 3.7, so that Symfony's DebugClassLoader can warn about the return type to prepare the migration.

odolbeau: Fixed in #214

odolbeau:
@stof does it looks OK for you? If yes, I can mere #211 then this issue. :)

PR #215, opened 2020-01-25
Add ServicesResetterProcessor (closes #193)

### LINKED ISSUE #193: Leverage ResetInterface
Symfony propose a https://github.com/symfony/contracts/blob/master/Service/ResetInterface.php
I think we could leverage this interface.

I'm thinking about doctrine and monolog.

For exemple, in monolog bundle, we clear eveything on kernel.reset.
Doctrine does the same BTW.

So I think we need to add a dependency on the event dispatcher to be able to dispatch a kernel.reset event after each consumed message.

Thanks to that, we could depreciate some Processors :)

WDYT?

pvgnd:


pvgnd:
@odolbeau could you have a look please ?

odolbeau:
Looks good to me, thanks @pvgnd 
@lyrixx WDYT?

PR #217, opened 2020-02-05
Correct publishers constructor

odolbeau:


phpstan.neon.dist
```diff
@@ -17,3 +17,13 @@ parameters:
             message: "#^Else branch is unreachable because ternary operator condition is always true\\.$#"
             count: 1
             path: src/Swarrot/Processor/Doctrine/ConnectionProcessor.php
+
+        -
+            message: "#^Method Swarrot\\\\Broker\\\\MessagePublisher\\\\PeclPackageMessagePublisher\\:\\:__construct\\(\\) has parameter \\$timeout with no typehint specified\\.$#"
```
stof, line 22: instead of ignoring them, some phpdoc should be added to document the type.

PR #222, opened 2020-09-30
Update processor for doctrine/persistence:^2.0

lepiaf:
Hello,

I update processor according change from swarrot bundle https://github.com/swarrot/SwarrotBundle/pull/202

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
     "require-dev": {
         "phpunit/phpunit": "^8.4",
         "php-amqplib/php-amqplib": "^2.1",
-        "doctrine/common": "^2.3",
+        "doctrine/common": "^2.9",
         "doctrine/dbal": "^2.4",
         "php-http/curl-client": "^2.0",
         "symfony/phpunit-bridge": "^5.0",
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -18,8 +18,8 @@
     "require-dev": {
         "phpunit/phpunit": "^8.4",
         "php-amqplib/php-amqplib": "^2.1",
-        "doctrine/common": "^2.3",
-        "doctrine/dbal": "^2.4",
+        "doctrine/common": "^3.0",
```
stof, line 21: this should not force the 3.0 version IMO, as the latest 2.x versions of common are still compatible. This should be using a `||` in the constraint.

stof (changes requested):

mm-pvgnd:
@odolbeau I though this MR has been already merged as the one in SwarrotBundle was. Could you merge & release please? Because since we released SwarrotBundle 2.1.0, this change in swarrot is missing.

composer.json
mm-pvgnd: @lepiaf has changed this to 2.9, maybe we can allow 3.0 in Swarrot & SwarrotBundle in other MRs

odolbeau: @stof is it OK for you?

PR #227, opened 2021-01-08
Document possible redeliveries

greg0ire:
We have been observing lots of redeliveries in our production
environment, and the only way I found to reproduce them was to use
`sleep($max_execution_time + 1)` at the beginning of my business
processor.
Increasing the max execution time resulted in a greater gap between
redeliveries, which confirms the hypothesis that this processor is
responsible for that.

The red bar on the graph below is the deploy success event for our application.

Screenshot_2021-01-08  SEARCH  Microservice Datadog

I was quite concerned because redeliveries, because of [how they are documented](https://www.rabbitmq.com/reliability.html#consumer-side):

> In the event of network failure (or a node failure), messages can be redelivered, and consumers must be prepared to handle deliveries they have seen in the past. It is recommended that consumer implementation is designed to be idempotent rather than to explicitly perform deduplication.

I thought something was wrong with our infrastructure, and asked around on the enterprise Slack if redeliveries were a bad thing, and was told that they probably were.

I think they might be avoided by doing the following:

```diff
     public function process(Message $message, array $options): bool
     {
         if (true === $this->isTimeExceeded($options)) {
             return false;
         }
 
         return $this->processor->process($message, $options);
+
+        if (true === $this->isTimeExceeded($options)) {
+            return false;
+        }
     }
```

That would probably result in the current message not being ack'ed right?

### DIFF
--- a/src/Swarrot/Processor/MaxExecutionTime/README.md
+++ b/src/Swarrot/Processor/MaxExecutionTime/README.md
@@ -3,3 +3,6 @@
 MaxExecutionTimeProcessor is a [swarrot](https://github.com/swarrot/swarrot)
 processor.
 Its goal is to stop the consumer when the max execution time have been reached.
+That might result in redeliveries in some cases, for instance if you start
+processing a message before the maximum execution time, then the next message
+will be redelivered.


stof:
> That would probably result in the current message not being ack'ed right?

depends on the order of your processors in the stack. If that processor is outside the AckProcessor, things should be OK.

stof:
note that your suggested patch does not solve the issue as is: it only adds dead code...

greg0ire:
Oh right, one would have to do a logical and with the line above:

`return $this->processor->process($message, $options) && !$this->isTimeExceeded($options)`.

> If that processor is outside the AckProcessor, things should be OK.

Indeed, and that's the recommended way to do things, so it might make sense to contribute the above then?

odolbeau:
Hi @greg0ire!

I agree with your latest proposal!
The child processor should be called even if the time is exceeded. as the message has been pulled from the broker anyway.

Nice catch! :+1:

greg0ire:
Great! Will contribute that!

PR #228, opened 2021-01-13
Check time after running the child processor

greg0ire:
There are two checks in the processor: on in process(), one in sleep().
In the event there are many messages in the queue when the time limit is
reached, it will be the check in process() that will kick in. Most users
will put the AckProcessor inside this one because that is what is
recommended in the docs, and will there will be no ack or nack in that
case, because of that check, which means there will be redeliveries.
By performing the check after running the child processor, we avoid the
redeliveries.

Closes #227

### DIFF
--- a/src/Swarrot/Processor/MaxExecutionTime/MaxExecutionTimeProcessor.php
+++ b/src/Swarrot/Processor/MaxExecutionTime/MaxExecutionTimeProcessor.php
@@ -61,11 +61,7 @@ class MaxExecutionTimeProcessor implements ConfigurableInterface, InitializableI
      */
     public function process(Message $message, array $options): bool
     {
-        if (true === $this->isTimeExceeded($options)) {
-            return false;
-        }
-
-        return $this->processor->process($message, $options);
+        return $this->processor->process($message, $options) && !$this->isTimeExceeded($options);
     }
 
     protected function isTimeExceeded(array $options): bool
--- a/tests/Swarrot/Processor/MaxExecutionTime/MaxExecutionTimeProcessorTest.php
+++ b/tests/Swarrot/Processor/MaxExecutionTime/MaxExecutionTimeProcessorTest.php
@@ -54,4 +54,27 @@ class MaxExecutionTimeProcessorTest extends TestCase
 
         $this->assertTrue(microtime(true) - $startTime > $maxExecutionTime);
     }
+
+    public function test_it_stops_after_slow_processing()
+    {
+        $message = new Message('body', [], 1);
+
+        $maxExecutionTime = 1;
+        $innerProcessor = $this->prophesize(ProcessorInterface::class);
+        $innerProcessor->process($message, ['max_execution_time' => $maxExecutionTime])->willReturn(true);
+
+        $logger = $this->prophesize(LoggerInterface::class);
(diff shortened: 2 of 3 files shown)


src/Swarrot/Processor/MaxExecutionTime/MaxExecutionTimeProcessor.php
```diff
@@ -61,11 +61,7 @@ public function sleep(array $options): bool
      */
     public function process(Message $message, array $options): bool
     {
-        if (true === $this->isTimeExceeded($options)) {
-            return false;
-        }
```
greg0ire, line 66: At first, I had not planned to remove this check, but after reading "The child processor should be called even if the time is exceeded. " from https://github.com/swarrot/swarrot/pull/227#issuecomment-758773279, I think it's better to remove it, because that check would only ever kick in if Swarrot itself was consuming the time, making redeliveries even harder to troubleshoot.

odolbeau: IMO, as soon as a message have been fetched from the broker, but must be processed.
The current implementation is not OK as this processor it will stop the swarrot consumer if the max execution time have been reached **even if the current message have not been consumed yet**.

That's why, the correct implementation should be:
1. process the message
2. check if time is exceeded
3. If it is, return false, otherwise, return the inner processor result

The check must not be removed, it must be moved **after** processing the current message and this is exactly what your current implementation do. :ok_hand:

tests/Swarrot/Processor/MaxExecutionTime/MaxExecutionTimeProcessorTest.php
```diff
@@ -54,4 +54,36 @@ public function test_count_default_messages_processed()
...
+        $numberOfProcessedMessages = 0;
+        while (true) {
+            $numberOfProcessedMessages++;
+            if (false === $processor->process($message, ['max_execution_time' => $maxExecutionTime])) {
+                break;
+            }
+        }
+
+        $this->assertSame(1, $numberOfProcessedMessages);
+    }
```
odolbeau, line 88:
```suggestion
        $this->assertFalse($processor->process($message, ['max_execution_time' => $maxExecutionTime]))
```

I think this sould be enough?

src/Swarrot/Processor/MaxExecutionTime/MaxExecutionTimeProcessor.php
greg0ire: It definitely sounds like the best solution 👍

odolbeau:
Thanks for your contribution! :ok_hand:

stof:
@greg0ire have you checked whether the processor limiting the number of messages (rather than the time) has the same issue ?

greg0ire:
@stof I didn't, but it looks good to me:

https://github.com/swarrot/swarrot/blob/cfb0ff0bf96f47659e8cfb1846ec205f9cff5885/src/Swarrot/Processor/MaxMessages/MaxMessagesProcessor.php#L33-L47

The check is performed after running the child processor, which means the ACK should be sent.

greg0ire:
Could you please tag a release @odolbeau ?

greg0ire:
Thank you!

PR #235, opened 2021-12-22
Move pipeline from Travis to Github actions

### LINKED ISSUE #234: CI needs to migrate away from Travis
The swarrot organization ran out of Travis credits due to the new Travis pricing, so we need to migrate the CI setup (I suggest using GitHub Actions)

metfan:
Closes #234


.github/workflows/unit.yml
```diff
@@ -0,0 +1,55 @@
...
+                php: [7.4, 8.0, 8.1]
+                symfony: [4.4.*, 5.4.*, 6.0.*]
+                exclude:
+                    - php: 7.4
+                      symfony: 6.0.*
+
+        services:
+            rabbitmq:
+                # 3.8.10 and later versions has enabled TLSv1.3 by default which is not supported by PHP images
+                # TLSv1.3 can be disabled by using configuration file, but not environment variables
```
stof, line 21: This comment looks weird to me. Is 3.8.10 disabling support for TLSv1.2 (supporting *only* 1.3) ? Or is the rabbitmq server or PHP shipping with a broken implementation of TLS negotiation ?

.github/workflows/unit.yml
```diff
@@ -0,0 +1,57 @@
...
+
+            - name: Start broker service
+              run: docker restart ${{ job.services.rabbitmq.id }}
+
+            - name: Install Symfony Flex
+              run: composer global require --no-progress --no-scripts --no-plugins symfony/flex
+
+            - name: Install the dependencies
+              env:
+                  SYMFONY_REQUIRE: ${{ matrix.symfony }}
```
stof, line 50: To me, we should have some CI jobs running with unmodified constraints, to ensure that our requirements are OK.

.github/workflows/unit.yml
```diff
@@ -0,0 +1,57 @@
+on: [push, pull_request]
+name: Unit test
+jobs:
+    tests:
+        name: PHP ${{ matrix.php }} / SF ^${{ matrix.symfony }}
```
stof, line 5: this `^` should be removed. `^6.0.*` is an invalid constraint.

.github/workflows/unit.yml
```diff
@@ -0,0 +1,57 @@
...
+                    - 15672:15672
+                env:
+                    RABBITMQ_DEFAULT_VHOST: swarrot
+
+        steps:
+            - name: Checkout
+              uses: actions/checkout@v2
+
+            - name: Setup PHP
+              uses: shivammathur/setup-php@2.16.0
```
stof, line 36: why pinning an exact version here and not for `actions/checkout` ?
Also, do you plan to use dependabot to maintain those pinned versions ?

.github/workflows/unit.yml
metfan: You right, sorry it's a copy/paste error.

.github/workflows/unit.yml
metfan: To be honest, I'm discovering github action and it's mainly copy/paste from other config I found. I removed these pinned versions.
Thank you for the heads-up

.github/workflows/unit.yml
stof: copy-paste from where ?

Also, if you remove the comment explaining why 3.8.10 is not used without actually using new versions, this makes the situation even worse IMO.

metfan: I just tryed with latest rabbitmq version 3.9 and it's working. 
Sorry I don't remember from where I pasted this part, I digged in so many repositories...

PR #241, opened 2023-01-03
Add $options types

yguedidi:


stof:
Please also add the same types in InitializableInterface, SleepyInterface, TerminableInterface and Consumer for other places using the options in method signatures.

stof (approved):

PR #245, opened 2023-07-03
Add DBAL3 support on ConnectionProcessor

rflavien:


src/Swarrot/Processor/Doctrine/ConnectionProcessor.php
```diff
@@ -68,16 +69,23 @@ public function process(Message $message, array $options): bool
...
                         $connection->close();
                     }
                 }
             }
         }
     }
 
+    private function isConnectedToPrimary(Connection $connection): bool
+    {
+        if (!class_exists('\Doctrine\DBAL\Connections\PrimaryReadReplicaConnection')) {
```
stof, line 82: this is not needed. You can totally do several instanceof checks (and this will be more efficient than triggering the autoloading with a `class_exists` check)

PR #247, opened 2023-09-13
Update the branch alias

stof:
This updates the branch alias to account for the renaming of master to main. It also updates the alias to avoid specifying a particular minor version, so that it does not need to be updated after each release (which had been forgotten)


odolbeau (approved):
Thanks! :ok_hand:

PR #248, opened 2023-09-13
Add a processor to reset the peak memory usage

stof:
This is useful when you have another processor that reports the peak memory usage and you want to distinguish it per message instead of depending on the history of previous messages as well.
For instance, I have a custom processor using the [tideways](https://tideways.com/) API to report profiles for the message processing.
The expected usage is that we would use this MemoryResetProcessor on the outside of the stack of processors, so that the peak memory usage is reset after it has been reported to Tideways (and also after the Doctrine EntityManager has been cleared for instance, to release memory).

To make it easier to write projects supporting multiple PHP versions, I made this processor a no-op on PHP 8.1 and older instead of making it unsupported.

### DIFF
--- a/src/Swarrot/Processor/MemoryReset/MemoryResetProcessor.php
+++ b/src/Swarrot/Processor/MemoryReset/MemoryResetProcessor.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace Swarrot\Processor\MemoryReset;
+
+use Swarrot\Broker\Message;
+use Swarrot\Processor\ProcessorInterface;
+
+class MemoryResetProcessor implements ProcessorInterface
+{
+    private ProcessorInterface $processor;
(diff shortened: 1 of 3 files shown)


stof:
To fix the phpstan job, https://github.com/swarrot/phpqa should be republished as its parent image has been updated to use PHP 8.2 now.

To simplify the setup, I would suggest to drop the usage of custom docker images to run the CI, defining the setup in the GHA workflows directly, which would allow the config to be managed in the repository directly instead of having it depend on external repositories.

stof:
@odolbeau what do you think about this proposal for the CI setup ? If you agree with this, I can do the change.

odolbeau:
Sure, looks like a great idea! :ok_hand: 
Thanks for this!

PR #249, opened 2023-09-19
Update CI setup

stof:
As discussed in https://github.com/swarrot/swarrot/pull/248#issuecomment-1717387397, this moves php-cs-fixer and phpstan to be managed by composer instead of the swarrot-phpqa image maintained in https://github.com/swarrot/phpqa (this package should be archived now).

Thanks to this change, we are now using up-to-date versions of PHP-CS-Fixer (which was using v2 before) and phpstan.
I also took this opportunity to update PHPUnit to 9.6 (I haven't updated to 10.x because we use the deprecation tooling of symfony/phpunit-bridge which is not compatible yet)

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -6,11 +6,20 @@ jobs:
   php-cs-fixer:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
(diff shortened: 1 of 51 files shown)


odolbeau (approved):
Thanks for this amazing work! :ok_hand:

PR #265, opened 2024-12-16
We need to talk about messenger

odolbeau:
If swarrot was a great choice a few years ago, I'm convinced that messenger as much more to offer than swarrot nowadays.

This PR add an alert on top of the README to invite readers to use the Symfony Component instead of this lib.

WDYT @swarrot/maintainers ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,9 @@
 [![Latest Stable Version](https://poser.pugx.org/swarrot/swarrot/v/stable.svg)](https://packagist.org/packages/swarrot/swarrot)
 [![Latest Unstable Version](https://poser.pugx.org/swarrot/swarrot/v/unstable.svg)](https://packagist.org/packages/swarrot/swarrot)
 
+> [!IMPORTANT]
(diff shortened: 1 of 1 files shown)


stof:
I'll keep maintaining swarrot for now (probably in feature-freeze mode though) as I'm still using it at work where we have our own messenger-like internal API built on top of it (but not easy to migrate to Messenger as the API is not the same, and we rely on some RabbitMQ routing that would not be easy to implement in Messenger).

But I'm fine with recommending to look at symfony/messenger for projects that don't have such needs.

odolbeau:
For sure, the goal is not to deprecate Swarrot as it's still heavily used. :ok_hand:

metfan:
I don't think Messenger should be the only option. 
I like to use RabbitMQ with a PHP lib and not use a PHP lib that accidentally use RabbitMQ. I mean some RabbitMQ capacity can't be used or at least only in kind of automatic thing with Messenger. Messenger is also a complex lib with not so easy extension when adding a processor in Swarrot is a piece of cake.

PR #266, opened 2024-12-17
Update PHP-CS-Fixer & PHPStan

odolbeau:


stof:
For phpstan, should we migrate to phpstan 2 ?

stof (commented):
would be great to also fix the CI on PHP 7.4 (or to bump the min PHP version)

composer.json
```diff
@@ -11,19 +11,19 @@
...
         "doctrine/dbal": "^3.0 || ^4.0",
         "php-amqplib/php-amqplib": "^2.1 || ^3.0",
         "phpspec/prophecy": "^1.15",
         "phpspec/prophecy-phpunit": "^2.0",
-        "phpunit/phpunit": "^9.6 || ^10.0",
-        "symfony/error-handler": "^5.4 || ^6.0 || ^7.0",
-        "symfony/phpunit-bridge": "^5.4 || ^6.0 || ^7.0",
+        "phpunit/phpunit": "^11.5",
+        "symfony/error-handler": "^6.4 || ^7.1",
+        "symfony/phpunit-bridge": "^6.4 || ^7.1",
```
stof, line 26: using the `symfony/phpunit-bridge` with PHPUnit 11.5 does not make sense, given that we use it for its deprecation reporting feature (we should use the builtin feature in PHPUnit 11 instead)

tests/Swarrot/Broker/MessageProvider/CallbackMessageProviderTest.php
```diff
@@ -84,9 +83,7 @@ public function __invoke()
         ];
     }
 
-    /**
-     * @dataProvider callable_without_message_provider
-     */
+    #[DataProvider('callable_without_message_provider')]
     public function test_get_without_messages_in_queue_return_null($callableProvider)
```
stof, line 87: you need to make data providers static.

composer.json
stof: We could keep it to make it register the DebugClassLoader of `symfony/error-handler` to trigger its deprecations, but this requires using version 7.2+ of that component (and registering the extension in the phpunit.xml)

odolbeau:
> For phpstan, should we migrate to phpstan 2 ?

Sure, this can be done in a separate PR

tests/Swarrot/Broker/MessageProvider/CallbackMessageProviderTest.php
stof: ah sorry, I messed up

phpunit.xml.dist
```diff
@@ -1,38 +1,20 @@
...
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" backupGlobals="false" colors="true" processIsolation="false" stopOnFailure="false" bootstrap="vendor/autoload.php" xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd" cacheDirectory=".phpunit.cache" backupStaticProperties="false">
+  <php>
+    <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[total]=999"/>
+  </php>
+  <testsuites>
+    <testsuite name="Tests">
+      <directory>./tests</directory>
+    </testsuite>
+  </testsuites>
+  <source>
```
stof, line 11: this does not configure the deprecation reporting following the Symfony convention of triggering them, so this is not a replacement for the phpunit-bridge for now.

phpunit.xml.dist
```diff
@@ -1,38 +1,20 @@
...
-    </php>
-
-    <testsuites>
-        <testsuite name="Tests">
-            <directory>./tests</directory>
-        </testsuite>
-    </testsuites>
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" backupGlobals="false" colors="true" processIsolation="false" stopOnFailure="false" bootstrap="vendor/autoload.php" xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd" cacheDirectory=".phpunit.cache" backupStaticProperties="false">
+  <php>
+    <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[total]=999"/>
```
stof, line 4: this does not make sense anymore if you don't use the bridge anymore for deprecations.

odolbeau:
@stof I don't understand which phpunit configuration you expect exactly? :thinking: 
I updated phpunit options in order to fail on deprecations but I'm not sure that's what you were looking for?
As far as I understand, phpunit-bridge is not overriding phpunit error handling system, that's why we don't need the bridge anymore in this lib right?

odolbeau:
@stof I merge this PR to repair the CI but don't hesitate to answer to my previous message, I'll improve the configuration following your advices. :)

PR #267, opened 2025-01-06
Improve connection with pecl

AurelienPillevesse:
Improve connection when using ext-amqp. Already implemented and explained in `symfony/messenger`.

```
Try to reconnect once to accommodate need for one of the nodes in cluster needing to stop serving the traffic. 
This may happen for example when one of the nodes in cluster is going into maintenance node.

See https://github.com/php-amqplib/php-amqplib/issues/1161
```

https://github.com/symfony/amqp-messenger/blob/7.2/Transport/AmqpReceiver.php#L53-L67

### DIFF
--- a/src/Swarrot/Broker/MessageProvider/PeclPackageMessageProvider.php
+++ b/src/Swarrot/Broker/MessageProvider/PeclPackageMessageProvider.php
@@ -18,7 +18,12 @@ class PeclPackageMessageProvider implements MessageProviderInterface
 
     public function get(): ?Message
     {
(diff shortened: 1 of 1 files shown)


odolbeau:
Thanks for your contribution!
Could you please add a comment to explain why this code is made please? Maybe you can copy / paste the comment in Symfony codebase. :)

CI have been fixed in #266, you can rebase your PR. :ok_hand:

odolbeau:
Great, thanks!

PR #269, opened 2025-01-13
Improve connection with pecl (part 2)

AurelienPillevesse:
Apply same fix as https://github.com/swarrot/swarrot/pull/267 for `ack` (https://github.com/php-amqp/php-amqp/blob/latest/stubs/AMQPQueue.php#L50) and `nack`(https://github.com/php-amqp/php-amqp/blob/latest/stubs/AMQPQueue.php#L238)  actions because `\AMQPConnectionException` can also happen here


odolbeau (approved):

odolbeau:
Thanks! :ok_hand:

